PR #25, opened 2024-04-18
Make lower and upper fields in Bound message optional

Shitaro:
This pull request addresses the limitation of the current `Bound` message in expressing decision variables without a lower or upper bound. To enhance the expressiveness of the protocol buffer schema and handle decision variables with unbounded lower (i.e., lower = $-\infty$) or upper (i.e., upper = $+\infty$) bounds, the following changes are proposed:

1. Make the `lower` and `upper` fields in the `Bound` message optional:

```protobuf
message Bound {
  // Lower bound of the decision variable.
  optional double lower = 1;
  // Upper bound of the decision variable.
  optional double upper = 2;
}
```

2. Update the `DecisionVariable` message to use the modified `Bound` message:

```protobuf
message DecisionVariable {
  // ...
  Bound bound = 3;
}
```

By making the `lower` and `upper` fields optional, we can now express decision variables with unbounded lower or upper bounds by omitting the corresponding field. This change improves the flexibility of the protocol buffer schema.

The proposed modifications have been implemented in this pull request. Please review the changes and provide feedback. If there are any concerns or suggestions for improvement, please let me know.


termoshtt:
I think the implementation of current main branch is better because following reasons:

- Considering the case of binary decision variables, the `bound` should be absent, and thus it is natural that the `bound` field of `DecisionVariable` is optional.
- Since `double` of protobuf can contain any IEEE-754 floating point value including $+\infty$ and $-\infty$, it is better to set `upper` by $+\infty$ explictly rather than setting `upper = None`. Especially, some people regards `lower = None` as $-\infty$ and others as $0$ because some format (e.g. MPS file) assumes lower bound is $0$ if omitted. We should avoid this ambiguity.

Shitaro:
Thank you for your perceptive comments and for taking the time to provide a comprehensive rationale for your perspective. Upon further reflection, I find myself in agreement with your assessment that the current implementation in the main branch is indeed the more suitable approach, given the compelling reasons you have outlined. Consequently, I have made the decision to close this pull request without merging the proposed changes.

Issue #117, opened 2024-09-27
typo `raw -> row`


yuyamashiro:
raw -> row

https://github.com/Jij-Inc/ommx/blob/260f6a83983bb8aade38a986e815682498431fd8/python/ommx/ommx/v1/__init__.py#L642

Shitaro:
Thank you for reporting the bug.
This is a very minor typo, so we'll fix it right away.

termoshtt:
Fixed in #119

PR #119, opened 2024-10-03
Fix the argument of `Quadratic`: `raws` -> `rows`

NY57:
Issue: https://github.com/Jij-Inc/ommx/issues/117


termoshtt:
Please update the dependency https://github.com/Jij-Inc/ommx/blob/260f6a83983bb8aade38a986e815682498431fd8/python/ommx-python-mip-adapter/pyproject.toml#L26

termoshtt (approved):
LGTM

PR #123, opened 2024-10-08
Add MPS file parser

Zengor:
Adds the ability to read MPS files to create OMMX instances.

This implementation is based on our internal parser used in jijmodeling. Some things could probably be changed to make the code simpler as they were mostly choices to make it convenient/fast to create JijModeling models specifically (eg. how the parser handles keeping track of variable and constraint types), but nothing major was changed to get the initial implementation done quicker.

This PR at the moment only includes the Rust interface for it, I'll be commiting the python side with some tests later, but the code should be reviewable at this point.

This adds flate2 and indoc as dependencies (indoc is only a dev-dependency). flate2 is required to read files from miplib directly. indoc is only for a minor test so let me know if it should be rewritten to avoid the dependency


rust/ommx/src/mps.rs
```diff
@@ -0,0 +1,90 @@
...
+//!
+
+use std::path::Path;
+
+mod convert;
+mod parser;
+
+use parser::*;
+
+pub fn load_mps_file(path: impl AsRef<Path>) -> Result<crate::v1::Instance, MpsParseError> {
```
termoshtt, line 60:
```suggestion
pub fn load_file(path: impl AsRef<Path>) -> Result<crate::v1::Instance, MpsParseError> {
```

Since this is in public `ommx::mps` module, "mps" in `load_mps_file` is redundant.

termoshtt:
Thanks!

PR #127, opened 2024-10-11
MPS File Reading Python API & Rust fixes

Zengor:
- Adds a python API for reading MPS files and an accompanying test
  - This is available as a function `ommx.mps.load_file` as well as a static method in `ommx.v1.Instance.load_mps`. 
- Adds a Rust API for loading an MPS file and automatically serializing to bytes (instead of as an `ommx::v1::Instance`)
  - Added mostly for convenience of implementation of the python API -- seems like doing this entirely in the pyo3 crate would require importing prost.
- Fixes a bug in the Rust parser where the objective function was also being added as an "empty" constraint.
- I also made it so we to start numbering IDs from 1 (instead of 0) when reading MPS files. This was mostly because any `id: 0` field doesn't get printed out with protobuf messages, but I'm not sure if we should just start from 0 anyway.


python/ommx/src/mps.rs
```diff
@@ -0,0 +1,15 @@
+use pyo3::{
+    exceptions::PyValueError,
+    prelude::*,
+    types::{PyBytes, PyString},
+};
+
+#[pyfunction(name = "load_mps_bytes")]
```
termoshtt, line 7:
```suggestion
#[cfg_attr(feature = "stub_gen", pyo3_stub_gen::derive::gen_stub_pyfunction)]
#[pyfunction(name = "load_mps_bytes")]
```

termoshtt:
> I also made it so we to start numbering IDs from 1 (instead of 0) when reading MPS files. This was mostly because any id: 0 field doesn't get printed out with protobuf messages, but I'm not sure if we should just start from 0 anyway.

IMO, this will cause future inconsistency, i.e. some loader returns 1-based ID and others does 0-based one. While this is acceptable, it would be better to fix the pretty printer of protobuf message rather changing internal data format.

I am planing to replace the `raw` interface in Python SDK generated by `protoc` with exposing Rust SDK's ones. This will replace the pretty printer by Rust's `Debug` or manually written better `Display` trait.

Please keep 0-based ID in this PR.

PR #137, opened 2024-10-23
Add MPS file output

Zengor:
https://www.notion.so/jijinc/OMMX-MPS-c0bd6737ffd142df80f805bdba7873d4

Adds the ability to write out OMMX instances as MPS files.
Files are emitted compressed with gzip, matching our parser's assumption that all MPS files are compressed.

Not all valid OMMX instances are supported. Only linear functions are accepted.


termoshtt:
Could you add a test for

1. Create an instance acceptable to MPS writer
2. Dump to MPS file
3. Load it and compare to the original one

Zengor:
Added a simple test

termoshtt (approved):
LGTM

PR #143, opened 2024-11-12
QUBO and PUBO format output

termoshtt:
- This PR adds methods `to_qubo` and `to_pubo` to create Quadratic Unconstrained Binary Optimization (QUBO) and Polynomial ... (PUBO) formats.
- PUBO contains arbitrary order of polynomial, i.e. QUBO is also a valid PUBO. Since the maximum order may be linear or quadratic, we use PUBO instead of HUBO (higher-order ...).


termoshtt:
Recreated with `ParametricInstance` by #146

PR #145, opened 2024-11-18
Add `model_to_solution` and `model_to_state`.

NY57:
# 変更点
- 変更前の時点では、`model_to_solution` 関数が`ommx.v1.State`を返している状態だったため、関数名を`model_to_state`に変更しました
- `ommx.v1.Solution`を返す関数`model_to_solution`を新しく実装しました
- この変更に伴い、`solve`関数の内部処理を新規実装した`model_to_solution`に置き換えました

# 注意事項
`model_to_solution`関数の破壊的変更のため、最低でもマイナーバーションを上げる必要があります。

### DIFF
--- a/python/ommx-python-mip-adapter/ommx_python_mip_adapter/python_mip_to_ommx.py
+++ b/python/ommx-python-mip-adapter/ommx_python_mip_adapter/python_mip_to_ommx.py
@@ -180,8 +180,8 @@ def model_to_state(
         >>> model.optimize()
         <OptimizationStatus.OPTIMAL: 0>
 
-        >>> ommx_solutions = adapter.model_to_state(model, ommx_instance)
-        >>> ommx_solutions.entries
+        >>> ommx_state = adapter.model_to_state(model, ommx_instance)
+        >>> ommx_state.entries
         {1: 0.0}
     """
     if not (
@@ -213,7 +213,6 @@ def model_to_solution(
     .. doctest::
 
         >>> from ommx.v1 import Instance, DecisionVariable
-        >>> from ommx.v1.solution_pb2 import Optimality
         >>> from ommx_python_mip_adapter import instance_to_model, model_to_solution
 
         >>> p = [10, 13, 18, 31, 7, 15]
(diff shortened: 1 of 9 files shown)


python/ommx-python-mip-adapter/ommx_python_mip_adapter/python_mip_to_ommx.py
```diff
@@ -198,3 +198,55 @@ def model_to_solution(
...
+    """
+    Convert optimized Python-MIP model and ommx.v1.Instance to ommx.v1.Solution.
+
+    Examples
+    =========
+
+    .. doctest::
+
+        >>> from ommx.v1 import Instance, DecisionVariable
+        >>> from ommx.v1.solution_pb2 import Optimality
```
termoshtt, line 216:
```suggestion
```

この行は要らないはず

Shitaro (approved):
I believe there is no need to add `model_to_state` as a separate function for the following reasons:

- The `state` field of `ommx.v1.Solution` already contains `ommx.v1.State`.   
- To my knowledge, there is no use case that requires retrieving only `ommx.v1.State`.

However, the implementation of model_to_solution will use model_to_state. It would not be desirable to pay the cost of making it private.

For these reasons, I believe the current approach in this PR is appropriate.

termoshtt:
> To my knowledge, there is no use case that requires retrieving only ommx.v1.State.

Since evaluating into `ommx.v1.Solution` is somewhat heavy, it would be better to use `State` instead of `Soluiton` when Python-MIP adapter is used as a component of higher level algorithm. Basically `State` is for computer and `Solution` is for person.

Shitaro:
> > To my knowledge, there is no use case that requires retrieving only ommx.v1.State.
> 
> Since evaluating into `ommx.v1.Solution` is somewhat heavy, it would be better to use `State` instead of `Soluiton` when Python-MIP adapter is used as a component of higher level algorithm. Basically `State` is for computer and `Solution` is for person.

In fact, as you mentioned, from a computational point of view, if just getting the state is sufficient, then it's better to just return the state without any further computation. However, I personally can't think of a situation where you would use just the state without the solution when implementing advanced algorithms. Wouldn't you generally need not only the intermediate solution but also the associated optimal value when implementing algorithms? For example, in a simple example like branch-and-bound, we would use both the solution of the relaxed subproblem (according to `State`) and its optimal value.
Well, it's quite possible that I simply can't imagine such use cases due to my limited experience and knowledge of mathematical optimisation.

In any case, I have no intention of opposing the addition of a function that returns `State`.

Shitaro (changes requested):
As this PR introduces breaking changes, could you please update the minor version to 1.4.0?

termoshtt (approved):
LGTM

PR #162, opened 2024-11-26
MPS output is broken if same name variables exist

termoshtt:
Fixes problems caused by trying to create and then read MPS files using an instance that has multiple variables/constraints with the same name.

- When creating an MPS file, variables and constraints are now named based on their IDs. The name metadata in the OMMX instance is ignored.
- When reading an OMMX file, if all names match the new ID-based naming scheme, those IDs will be recovered and reused. Otherwise, the names are interpreted as metadata and IDs are assigned in an unspecified order, as before.


rust/ommx/src/mps/convert.rs
```diff
@@ -37,39 +51,72 @@ fn convert_dvars(mps: &Mps) -> (Vec<v1::DecisionVariable>, HashMap<ColumnName, u
...
-            ..Default::default()
-        })
+    // We want to be able to recover IDs if the var name is in the form
+    // "OMMX_VAR_<number>", matching how our output formats names.
+    //
+    // NOTE considering the case where an OMMX-created MPS file was later
+    // edited, there may be a mix of valid OMMX id names and invalid names.
+    // Handling all edge cases would be pretty complex and potentially bad for
+    // performance. For simplicity, we only apply ID recovery when ALL variables
+    // match the naming pattern.
```
termoshtt, line 61: This is good enough. Thanks!

termoshtt (commented):
LGTM

PR #172, opened 2024-11-28
Introduce graded-lexical order for monomial

termoshtt:
Split from #143


termoshtt:
This cannot sort user input in desired order. This feature serves no purpose.

PR #229, opened 2024-12-18
MPS: Only write used variables to output

Zengor:
As a lot of solvers error out if the `BOUNDS` section includes unused variables, this PR updates the MPS output writer to only add variables that were actually used in the problem to the `BOUNDS` section. 

In this PR I also added a generic `anyhow::Error`→`MpsWriteError` conversion asn "Unknown Error", but this is mostly just to make the code compile. The method `used_variable_ids()` returns an `anyhow::Result`, but it doesn't seem to be able to produce any errors at all. Still, changing that method would be an API change so I decided to leave it as is. I added the error conversion instead of unwrapping just in case.

### DIFF
--- a/rust/ommx/src/convert/instance.rs
+++ b/rust/ommx/src/convert/instance.rs
@@ -28,7 +28,7 @@ impl Instance {
         }
     }
 
-    pub fn used_decision_variable_ids(&self) -> Result<BTreeSet<u64>> {
+    pub fn used_decision_variable_ids(&self) -> BTreeSet<u64> {
         let mut used_ids = self.objective().used_decision_variable_ids();
         for c in &self.constraints {
             used_ids.extend(c.function().used_decision_variable_ids());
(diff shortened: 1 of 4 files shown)


termoshtt:
> Still, changing that method would be an API change so I decided to leave it as is. 

Since this PR will be released as OMMX Rust SDK 2.0.0, this breaking change can be acceptable.

> In this PR I also added a generic anyhow::Error→MpsWriteError conversion asn "Unknown Error", but this is mostly just to make the code compile. 

I propose to remove `MpsWriteError` by replacing to `anyhow::Error`. There is little reason to separate this error statically.

Zengor:
I changed the return type on `used_decision_variable_ids` to remove the `Result`.

Regarding `MpsWriteError`, I think there's value in having custom error types that list the possible errors that can happen, especially in a library like ommx. In this particular case it's also a bit convenient for our code as we check the error type in `write_columns` to switch to a message more appropriate to the objective function (though of course we could achieve that in other ways). If you insist I'll remove `MpsWriteError` but I've left it for now.

termoshtt:
> If you insist I'll remove MpsWriteError but I've left it for now.

Ok, no need to remove it. Please fix test failures.

Zengor:
Sorry, I forgot to run the tests locally. Fixed.

termoshtt:
Thanks a lot

PR #250, opened 2024-12-24
QPLIB format parser

Zengor:
Adds the ability to read QPLIB files to create OMMX instances.

Some consideration has been given to adding the ability to output QPLIB format in the future but some aspects of it are still open.


rust/ommx/src/qplib/convert.rs
termoshtt, line 1: Similar to MPS, we want to create some cases which is invariant via the round-trip ommx.v1.Instance -> QPLIB format -> ommx.v1.Instance. Do you have any idea especially for ID mapping.

termoshtt (changes requested):

rust/ommx/src/qplib/convert.rs
Zengor: I've been thinking about it and there are a few ways but also some unresolved issues.

The biggest problem is that the QPLIB is meant to reduce the number of constraints by allowing two-sided `c_l <= expr <= c_r` inequalities, and so to have shorter QPLIB files we'd need to add special handling for that when outputting. On the other hand that would be pretty annoying to implement as the expressions wouldn't be equal in OMMX (one of them would be multiplied by -1). It'd be simpler for our implementation to ignore this and just double up on constraints when necessary, especially if what we care about is mostly the stability of IDs.

The other big problem is just the fact that OMMX IDs don't need to be contiguous, while QPLIB format IDs do have that assumption (as it just lists the number of variables/constraints at the top and assumes they will be numbered 1..N and 1..M)

When the IDs in the OMMX instance are sequential _and_ starting from 1, we could just have a 1:1 correspondence between the file IDs and the instance IDs. We'd still need a metadata marker so the parser knows to recover IDs for constraints (as opposed to the current way the parser handles splitting constraint IDs). But this would always have to be just a special case if we choose to add this capability.

For the more general case, the QPLIB file does still allow metadata like variable/constraint names, which we can use to have a very similar setup to our MPS ID mapping: just give them special names that contains their OMMX IDs, while the IDs in the file are fully sequential and not required to follow the OMMX instance. This is probably the simplest to implement.

termoshtt: Great, it makes the problem clear.

It looks unrealistic to create a direct mapping from OMMX ID to QPLIB ID as you say. It would be better to use metadata to keep tracking the information of OMMX side in QPLIB format.

Anyway, this will be an issue after implementing QPLIB output. I will merge this PR as it is. Thanks a lot!

PR #257, opened 2024-12-24
Add `decision_variable_dependency` to represent polynomial dependency between decision variables

termoshtt:
Ready for binary encoding of integer variables.

- When an integer variable $x \in [0, 3]$ is binary encoded as $x = b_1 + 2 b_2$ with $b_1, b_2 \in \{0, 1\}$, $x$ will disappear from objective function and constraints.
- Thus, $x$ is not passed to solver which only solve with $b_1$ and $b_2$. So we have to reconstruct $x$ from $b_1$ and $b_2$, but how?
- `decision_variable_dependency` field in `ommx.v1.Instance` stores this relation, i.e. $x = b_1 + 2 b_2$ as a set of the ID of $x$ and a polynomial $b_1 + 2 b_2$.

### DIFF
--- a/proto/ommx/v1/instance.proto
+++ b/proto/ommx/v1/instance.proto
@@ -60,4 +60,7 @@ message Instance {
 
   // Constraints removed via preprocessing. These are restored when evaluated into `ommx.v1.Solution`.
   repeated RemovedConstraint removed_constraints = 8;
+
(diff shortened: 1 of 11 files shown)


termoshtt:
More comprehensive test for evaluate dependency will be done when implementing log-encoding.

PR #321, opened 2025-02-14
add ommx-gurobipy-adapter

yuyamashiro:
add ommx-gurobipy-adapter


Shitaro:
Thank you for creating a new adapter with OMMX.
We would like the adapter added to OMMX to be for OSS only.
Therefore, please create a separate repository for the Gurobi adapter.

PR #323, opened 2025-02-14
Revise ommx.v1.ParametricInstnace

termoshtt:
- Revised: https://ommx-en-book--323.org.readthedocs.build/en/323/user_guide/parametric_instance.html (press "d" to show diff)


Shitaro (approved):
LGTM

PR #332, opened 2025-02-25
Add ommx-highs-adapter

Zengor:


NY57 (changes requested):
I would like to ask you to make a few changes.

- Add the string "ommx-highs-adapter" to [this line](https://github.com/Jij-Inc/ommx/blob/a6492364143f24502cff6ac951b5b6000fec0800/python/set-version.py#L97).
- Resolve pyright errors.
- Suggestions below:

NY57 (changes requested):

NY57 (approved):
LGTM!

PR #337, opened 2025-02-28
Prepare Python SDK 1.8.4 release

Zengor:
- Adds ommx-highs-adapter to the release workflow
- Fixes as a small papercut in ommx-python-mip-adapter where the exception type was not being exported in `__all__`


NY57 (approved):
LGTM!

PR #361, opened 2025-03-13
Add ruamel-yaml to dev dependencies for LSP support

termoshtt:
Follow up of #356 
Since `ruamel` is installed only via the head `script` section of `ci-scripts/setup-python-test-project.py`, but `pylance` does not see it.


konn (approved):
Good catch, thanks!

PR #390, opened 2025-04-07
replace images in introduction

taqro:
In the English version of the document's introduction page, Japanese version images were being used.
Therefore, I replaced them with the following English versions:

- `introduction_01.png`
- `introduction_02.png`
- `introduction_03.png`


NY57 (approved):
LGTM!

PR #462, opened 2025-05-19
Set Google Analytics (GA4) tracking code in Jupyter book

taqro:
I set GA4.


j-i-k-o (approved):
Confirmed 
* `analytics` is set to both English and Japanese notebook
* Build Jupyter Book with `GA_TRACKING_ID` replaced.

PR #565, opened 2025-07-17
Drop `impl Evaluate` for `ommx::v1::*`

termoshtt:


termoshtt:
It is hard to completely remove `v1::*` here.

PR #573, opened 2025-07-24
Error types does not have `#[non_exhaustive]` attribute

termoshtt:


termoshtt:
Though this PR is incompatible to release 2.0.1, we regard this is an hot bug fix of 2.0.0.

PR #799, opened 2026-04-16
Add Propagate trait with unit propagation and Provenance tracking

termoshtt:
## Summary

Implement unit propagation for special constraint types (OneHot, SOS1, Indicator) via a new `Propagate` trait, with atomic `Instance::partial_evaluate`.

### `Propagate` trait

```rust
pub enum PropagateOutcome<T: Propagate> {
    /// Constraint remains active (possibly shrunk / modified).
    Active(T),
    /// Constraint is fully determined by the state. Move to removed set as-is.
    Consumed(T),
    /// Constraint transformed into another type (e.g. Indicator → Constraint).
    /// `original` goes to removed set, `new` is the replacement.
    Transformed { original: T, new: T::Transformed },
}

pub trait Propagate: Sized {
    type Transformed;
    fn propagate(
        self,
        state: &State,
        atol: ATol,
    ) -> Result<(PropagateOutcome<Self>, State)>;
}
```

`Propagate` consumes `self` and returns the outcome + additional variable fixings. Atomicity is the caller's responsibility — individual impls may lose `self` on error (see below for how `Instance` handles this).

### Per-constraint propagation

| Constraint | `Transformed` | Key behaviors |
|-----------|---------------|---------------|
| OneHotConstraint | `()` | var=1 → Consumed (fix rest to 0); all=0 → Err; exactly one unfixed → unit clause (Consumed, fix to 1); otherwise Active (shrunk) |
| Sos1Constraint | `()` | var≠0 → Consumed (fix rest to 0); all=0 → Consumed (vacuous); multiple≠0 → Err; otherwise Active (shrunk) |
| IndicatorConstraint | `IndicatorPromote` | indicator=1 → Transformed (promote inner to regular Constraint); indicator=0 → Consumed; otherwise Active (function partial-evaluated) |

### Atomic `Instance::partial_evaluate`

Since `Propagate::propagate` consumes `self` and loses it on error, Instance-level atomicity is guaranteed via the clone/commit pattern:

```rust
fn partial_evaluate(&mut self, state, atol) -> Result<()> {
    let mut working = self.clone();  // snapshot
    // all fallible operations on `working`:
    working.propagate_special_constraints(...)?;
    working.objective.partial_evaluate(...)?;
    // ...
    *self = working;  // commit atomically
    Ok(())
}
```

On any failure, `self` is unchanged.

### Conflict detection in propagation fixed-point

`merge_state` detects when two propagated fixings disagree (e.g. one says x=0, another x=1 within tolerance) and returns an error rather than silently overwriting.

### Provenance tracking

`ConstraintMetadata.provenance: Option<Provenance>` records that a constraint was promoted from an indicator constraint. Used when indicator=1 triggers the `Transformed` path.

### ID allocation

`ConstraintCollection::unused_id()` allocates a fresh ID (`max(existing id) + 1`, or `0` if empty) for promoted constraints so they don't collide with existing IDs.

### Strict indicator value validation

Indicator variable values are validated to be within tolerance of 0 or 1. Values like `2.0` no longer silently accepted as ON.

## Test plan

- [x] `cargo test -p ommx` — 462 unit tests + 53 doc tests pass
- [x] `cargo clippy -p ommx` — clean
- [x] `cargo fmt --check` — clean
- [x] Tests per constraint type: Active / Consumed / Transformed / error paths
- [x] Indicator: on-promotes, off-consumed, not-fixed partial-eval, on-with-function-partial-eval

### DIFF
--- a/python/ommx-pyscipopt-adapter/tests/test_sos1_partial_evaluate.py
+++ b/python/ommx-pyscipopt-adapter/tests/test_sos1_partial_evaluate.py
@@ -38,9 +38,6 @@ def sos1_instance_setup():
     )
 
 
-@pytest.mark.skip(
-    reason="SOS1 variable partial_evaluate is deferred to the forget phase"
-)
 def test_adapter_handles_sos1_variable_fixed_nonzero(sos1_instance_setup):
     """Test that PySCIPOpt adapter handles instances when SOS1 variable is fixed to non-zero value."""
     instance = sos1_instance_setup
@@ -64,9 +61,6 @@ def test_adapter_handles_sos1_variable_fixed_nonzero(sos1_instance_setup):
     assert fixed_var.value == 5.0, "Fixed variable should have the specified value"
 
 
-@pytest.mark.skip(
-    reason="SOS1 variable partial_evaluate is deferred to the forget phase"
-)
 def test_adapter_handles_sos1_variable_fixed_to_zero(sos1_instance_setup):
     """Test adapter behavior when SOS1 variable is fixed to zero."""
     instance = sos1_instance_setup
--- a/python/ommx/src/constraint.rs
+++ b/python/ommx/src/constraint.rs
@@ -106,6 +106,7 @@ impl Constraint {
                 subscripts,
                 parameters: parameters.into_iter().collect(),
                 description,
+                provenance: None,
             },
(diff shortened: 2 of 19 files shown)


termoshtt:
@codex review

PR #800, opened 2026-04-17
Add ommx-derive crate with #[derive(LogicalMemoryProfile)]

termoshtt:
## Summary

- Introduces a new internal `ommx-derive` proc-macro crate providing `#[derive(LogicalMemoryProfile)]` for named-field structs, eliminating the silent miss-a-field failure mode of the hand-maintained `impl_logical_memory_profile!` macro.
- Scopes the profiling infrastructure (`LogicalMemoryProfile` trait, `LogicalMemoryVisitor`, `Path`/`PathGuard`, declarative macro, free helpers) to `pub(crate)` — they are no longer part of the public API surface.
- Renames the internal `FoldedCollector` to a public `MemoryProfile` struct with `Display`, `entries()`, `total_bytes()`, `len()`, `is_empty()`, and exposes it as the return type of a new `Instance::logical_memory_profile()` method. The Python binding keeps its `str` return type by calling `.to_string()` on the result.
- Migrates six in-crate types from the declarative `impl_logical_memory_profile!` macro to `#[derive(LogicalMemoryProfile)]`. The migration immediately surfaced and fixed a real bug: the `Instance` invocation had been out of sync with the struct and was silently omitting the `named_functions` field from every memory profile.
- Folds the standalone design doc into rustdoc on the relevant items.

## Breaking changes

Everything under `ommx::logical_memory` except `MemoryProfile` is now `pub(crate)`. Any external code that imported `ommx::logical_memory::{LogicalMemoryProfile, LogicalMemoryVisitor, Path, PathGuard, logical_memory_to_folded, logical_total_bytes}` will no longer compile. The sole public entry point is `Instance::logical_memory_profile() -> MemoryProfile`; Python callers see no change (still `str`). This is a deliberate surface-area reduction — only the Python binding was ever a real consumer of the old public API, and it now goes through the new method. Acceptable at v3.0.0-alpha.

## Why

The pre-existing design required hand-written `LogicalMemoryProfile` impls for every OMMX type. Adding a new field to a struct silently skipped it from profiling because the declarative macro invocation has to be updated by hand — nothing failed to compile. A `#[derive]` closes that gap by reflecting the struct definition at compile time.

At the same time, only `Instance::logical_memory_profile()` was ever actually consumed externally (via the Python binding), so the trait / path / visitor types were public without benefit. This PR tightens them to `pub(crate)` and keeps `MemoryProfile` as the single stable return type.

## Notable implementation points

- `extern crate self as ommx;` is added in `rust/ommx/src/lib.rs` so the derive's generated `::ommx::logical_memory::*` paths resolve from inside the `ommx` crate.
- `ommx-derive` is published to crates.io (because `ommx` has a strict `=` version dependency on it — crates.io resolves that against its own index) but has no stable standalone API. The release workflow publishes `ommx-derive` immediately before `ommx` in lockstep, same pattern as `serde`/`serde_derive`. External consumers should not depend on `ommx-derive` directly.
- The `impl_logical_memory_profile! { path::to::Type as "Name" { ... } }` declarative form is retained for the two call sites that target external (prost-generated) types: `v1::Parameters` and `v1::instance::Description`. Derive cannot target types defined in a foreign module, so those intentionally remain on the declarative macro.
- `NamedFunction` and `NamedFunctionID` gain `LogicalMemoryProfile` impls (the former via derive, the latter as a leaf) to support the Instance migration. Instance snapshots are updated to include the one new line (`Instance.named_functions;BTreeMap[stack] 24`) — no other changes, confirming the derive's output matches the declarative macro for the remaining fields.

## Derive constraints (documented in `rust/ommx-derive/src/lib.rs`)

- Named-field structs only; tuple structs, unit structs, and enums emit a `compile_error!`.
- Generic parameters are propagated but no `T: LogicalMemoryProfile` bound is added automatically — callers write their own `where` clauses.
- No field-skip attribute; types that genuinely shouldn't enumerate every field should hand-write the impl.
- Frame names are always `"TypeName.field_name"`; to override the type name (e.g. for foreign types), use the declarative macro's `as "Name"` form.

## Tests

- `ommx-derive`: 6 inline `insta` snapshot tests locking in generated code for flat / single-field / empty / generic / enum-error / tuple-error inputs.
- `ommx`: existing snapshot tests (constraint / decision-variable / instance / polynomial / assignments) pass with the migration; Instance snapshots update by exactly one line each (`Instance.named_functions;...`) reflecting the bug fix.
- Derive-vs-declarative-macro equivalence test in `rust/ommx/src/logical_memory/tests.rs` asserts both produce identical output for the same field set.
- `cargo test -p ommx` (467 lib + 46 integration/doc) and `cargo test -p ommx-derive` pass.
- `cargo clippy -p ommx -p ommx-derive` clean.
- Python SDK: `task python:ommx:pytest-unit` (167 passed) and `task python:ommx:pytest-doctest` pass unchanged.

## Test plan

- [ ] CI: Rust tests
- [ ] CI: Python tests (unit + doctest)
- [ ] CI: clippy / fmt
- [ ] Reviewer spot-check: `Instance::logical_memory_profile()` round-trip (Rust struct → `to_string()` → Python `str`)
- [ ] Reviewer spot-check: confirm the `named_functions` line now appears in real-world profile output


termoshtt:
@codex review

PR #801, opened 2026-04-17
Port QUBO/HUBO and slack helpers to the domain Instance

termoshtt:
## Summary

Step 1 of moving operations off the `v1::*` protobuf types onto the domain `ommx::*` layer. The principle we are working toward is **`v1::*` is IO-only; all mathematical operations live on the domain types**.

This PR:
- Ports four helpers from `v1::Instance` onto `ommx::Instance`:
  - `as_qubo_format`
  - `as_hubo_format`
  - `convert_inequality_to_equality_with_integer_slack`
  - `add_integer_slack_to_inequality`
- Adds `Function::evaluate_bound(&Bounds) -> Bound` (needed by the slack helpers) as `rust/ommx/src/function/evaluate_bound.rs`.
- Rewires the Python bindings in `python/ommx/src/instance.rs` so the four methods call the domain implementations directly instead of round-tripping through `v1::Instance`.
- Fixes `Function::content_factor` for the `Function::Constant` case (`c.inv()` → `c.inv().abs()`) so it returns a positive factor like the `PolynomialBase` path. Without this fix, negative constant constraints would sign-flip when multiplied by the content factor inside the slack helpers.
- Rejects special constraints (Indicator/OneHot/SOS1) in `as_qubo_format` / `as_hubo_format` via `check_capabilities(&empty_set)`, matching the guard pattern already present in `penalty_method`. Without this guard an instance with only special active constraints would silently serialize to a QUBO/HUBO that dropped them. The `slack` helpers operate on the regular constraint ID space only and need no analogous guard.
- Removes the migrated helpers (`as_qubo_format`, `as_hubo_format`, `convert_inequality_to_equality_with_integer_slack`, `add_integer_slack_to_inequality`, `log_encode`) from `v1_ext/instance.rs` since Python no longer calls the v1 copies.

The rest of `v1_ext/` is left in place for now — it is still used internally by `random/`, `qplib/`, various `evaluate.rs` impls, and `solution/parse.rs`. Tracking issue for the remaining migration: #802.

## Commits

- `0f8b4ca9` — Port QUBO/HUBO and slack helpers to the domain Instance
- `05c9d75c` — Remove migrated QUBO/HUBO and slack helpers from v1_ext
- `5b6ad181` — Remove v1_ext::log_encode now that Python calls the domain version
- `781af8d7` — Validate slack coefficient before mutating the instance
- `379b64e6` — Reject special constraints in as_qubo_format / as_hubo_format

## Review

Reviewed by Codex twice (initial pass + follow-up after the atomicity fix). Both passes returned **approve**. Highlights verified against the v1_ext originals:

- Mathematical equivalence per-method.
- `content_factor` fix matches the "minimal positive factor" contract and aligns with `PolynomialBase::content_factor`.
- Cancellation logic in `as_qubo_format` / `as_hubo_format` preserved (distinct monomials can collapse to the same binary key; the entry + accumulate + remove sequence is kept).
- `add_integer_slack_to_inequality` no longer leaves an orphan slack decision variable when `Coefficient::try_from(b)` fails (e.g. `slack_upper_bound == 0` giving `b = inf`). Covered by a regression test.
- `convert_inequality_to_equality_with_integer_slack` has no analogous failure window after mutation — all fallible checks run before `new_decision_variable`.
- Special constraints are now rejected at the boundary; a OneHot-only instance is verified to produce an error from `as_qubo_format`.

## Test plan

- [x] `cargo test -p ommx --lib` — 479 passed (15 new tests across `instance/qubo.rs`, `instance/slack.rs`, `function/evaluate_bound.rs`)
- [x] `task python:ommx:test` — 37 passed
- [x] `task python:ommx-openjij-adapter:test` — 30 passed (exercises `as_qubo_format` / `as_hubo_format` / slack helpers end-to-end)
- [ ] CI green

Tracking issue for the remaining v1_ext migration work: #802

### DIFF
--- a/python/ommx/src/instance.rs
+++ b/python/ommx/src/instance.rs
@@ -4,7 +4,7 @@ use crate::{
     Rng, SampleSet, Samples, Sense, Solution, State, VariableBound,
 };
 use anyhow::Result;
-use ommx::{ConstraintID, Evaluate, NamedFunctionID, Parse, VariableID};
+use ommx::{ConstraintID, Evaluate, NamedFunctionID, VariableID};
 use pyo3::{
     exceptions::PyKeyError,
     prelude::*,
@@ -386,8 +386,7 @@ impl Instance {
     }
 
     pub fn as_qubo_format<'py>(&self, py: Python<'py>) -> Result<(Bound<'py, PyDict>, f64)> {
-        let inner: ommx::v1::Instance = self.inner.clone().into();
-        let (qubo, constant) = inner.as_qubo_format()?;
+        let (qubo, constant) = self.inner.as_qubo_format()?;
         Ok((
             serde_pyobject::to_pyobject(py, &qubo)?
                 .extract()
@@ -397,8 +396,7 @@ impl Instance {
     }
 
     pub fn as_hubo_format<'py>(&self, py: Python<'py>) -> Result<(Bound<'py, PyDict>, f64)> {
-        let inner: ommx::v1::Instance = self.inner.clone().into();
-        let (hubo, constant) = inner.as_hubo_format()?;
+        let (hubo, constant) = self.inner.as_hubo_format()?;
         Ok((
             serde_pyobject::to_pyobject(py, &hubo)?
                 .extract()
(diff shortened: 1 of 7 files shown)


termoshtt:
Addressed all four Copilot review comments in `ef909f24`:

- `slack.rs` convert/add: `slack_id` now derived from `slack.id()` on the returned `&mut DecisionVariable`, removing the implicit coupling to `new_decision_variable`'s internal `next_variable_id()` call.
- `slack.rs:38`: `convert_inequality_to_equality_with_integer_slack` now bails up front when the constraint's equality is not `LessThanOrEqualToZero`, matching the sibling method and the method's documented contract. Covered by a new `convert_inequality_rejects_equality_constraint` test.
- `function/evaluate_bound.rs:61`: test comment rewritten — `Bound::pow(2)` on a mixed-sign interval yields `[0, 9]` (sound interval-power semantics), not the naive `[-6, 9]`. Expected value updated accordingly.

Tests: `cargo test -p ommx --lib` → 480 passed (one new test added).

PR #809, opened 2026-04-20
Benchmark pyscipopt adapter: SOS1 vs big-M

termoshtt:
## Summary

Adds a Plant Placement Problem benchmark for `ommx-pyscipopt-adapter` that compares **eight** equivalent OMMX formulations of "at most one plant per region", spanning three orthogonal axes:

- **`δ + big-M`** — whether the binary opening indicator `δ_i ∈ {0, 1}` and the big-M link `c_i ≤ C_i · δ_i` are introduced.
- **`Σδ ≤ 1`** — whether the explicit linear cardinality bound on `δ` is added. *Note*: when SOS1 is also declared (on `c`, on `δ`, or both), this linear bound is **logically redundant** — SOS1 already implies it. So the `_with_card` row in each pair is testing whether SCIP gets value out of the redundant hint, or whether it just adds overhead. The exception is (8) `bigm`, which has no SOS1 — there `Σδ ≤ 1` is the only thing enforcing "at most one plant per region", not redundant.
- **SOS1 location** — on the continuous capacities `{c_i}`, on the binary indicators `{δ_i}`, on **both** (redundant), or absent.

The eight builders enumerate every well-defined combination:

|     | Builder | δ + big-M | `Σδ ≤ 1` | SOS1 on `c` | SOS1 on `δ` |
|---|---|:---:|:---:|:---:|:---:|
| (1) | `build_sos1` | – | – | ✓ | – |
| (2) | `build_sos1_on_c_with_delta` | ✓ | – | ✓ | – |
| (3) | `build_sos1_on_c_with_delta_with_card` | ✓ | ✓ | ✓ | – |
| (4) | `build_sos1_on_delta` | ✓ | – | – | ✓ |
| (5) | `build_sos1_on_delta_with_card` | ✓ | ✓ | – | ✓ |
| (6) | `build_sos1_on_both_with_delta` | ✓ | – | ✓ | ✓ |
| (7) | `build_sos1_on_both_with_delta_with_card` | ✓ | ✓ | ✓ | ✓ |
| (8) | `build_bigm` | ✓ | ✓ | – | – |

All eight have the same feasible region on `(s, c)` and the same optimum (verified by `test_placement_equivalence.py`).

A note on (2): `δ` is introduced but is *not* constrained by anything except the big-M link `c_i ≤ C_i · δ_i`, which only enforces "`δ_i = 0` ⇒ `c_i = 0`" (one direction). Multiple `δ_i` could remain `1` simultaneously; "at most one plant per region" is enforced *entirely* by SOS1 on `c`. So `δ` in (2) is essentially a "free spectator" variable. As an `(s, c)` optimization problem this is mathematically valid (objective doesn't depend on `δ`), but `δ` carries no semantic meaning. (2) functions as a **control case** isolating "does adding `δ` to the model alone slow SCIP down" from any other effect.

## What is measured

The benchmark measures **SCIP's own processing time** (presolve + B&B), isolated from the adapter:

- `Input.random` → `ommx.v1.Instance` → `pyscipopt.Model` is built in **session-scoped fixtures**, outside the measured region.
- Each benchmark calls `model.freeTransform()` (discards SCIP's transformed problem from any previous run) + `model.optimize()` on the pre-built models.

This avoids biasing the comparison by adapter conversion cost, which scales with variable/constraint count and is therefore *cheaper* for `build_sos1` (no `δ`, fewer constraints) than for the seven δ-bearing variants.

Since the measured time is essentially "how long SCIP takes" (the adapter is out of the loop), this benchmark is most useful as a **one-off comparison run locally** to understand how the formulation choice interacts with SCIP. It is **not** wired into `.github/workflows/bench.yml` — tracking it in the CodSpeed CI dashboard would mostly track SCIP's behavior across upstream changes, which is out of scope for this repo.

## Placement problem module (reusable)

Lives at **`ommx.testing.placement`** so other adapters (`ommx-highs-adapter`, `ommx-python-mip-adapter`, …) can reuse the same comparison. The module docstring documents the full mathematical formulation — sets, parameters, decision variables, constraints, objective, and the 8-row table above — rendered under KaTeX in the API Reference (AutoAPI picks up `ommx.testing.placement` automatically under the existing `ommx.testing` entry).

To keep existing imports `from ommx.testing import SingleFeasibleLPGenerator, DataType` working, `testing.py` was converted to a `testing/` package (contents moved to `testing/__init__.py` unchanged); `placement.py` is a sibling submodule.

## Files

- `python/ommx/ommx/testing/placement.py` — `Input/Plant/Client` dataclasses, the eight builders, a shared `_build_with_delta` backbone parameterised by the three axis flags, and the full problem spec in the module docstring.
- `python/ommx/ommx/testing/__init__.py` — renamed from the old `testing.py`, contents unchanged.
- `python/ommx-pyscipopt-adapter/tests/test_placement_equivalence.py` — verifies all eight builders reach the same optimum on a small instance.
- `python/ommx-pyscipopt-adapter/tests/test_bench_placement.py` — `@pytest.mark.benchmark` tests parameterised over problem sizes for each builder; measures pure SCIP time via fixture-prebuilt models + `freeTransform()` + `optimize()`.
- `python/ommx-pyscipopt-adapter/Taskfile.yml` — `pytest` now excludes `test_bench_*.py`; adds `pytest-bench` (walltime) and `bench` (`--codspeed`).

## Local codspeed measurement (pure SCIP time)

`task python:ommx-pyscipopt-adapter:bench` on a dev machine (3 instances per param set; walltime mode; total runtime ≈5.5 min):

| size | sos1 (1) | sos1_on_c_with_delta (2) | …with_card (3) | sos1_on_delta (4) | …with_card (5) | sos1_on_both (6) | …with_card (7) | bigm (8) |
|---|---|---|---|---|---|---|---|---|
|  6 × 10  |   1.8 ms |   1.8 ms |   4.5 ms |   4.8 ms |   5.2 ms |   4.6 ms |   4.6 ms |   5.1 ms |
| 12 × 20  |  17.5 ms |  17.5 ms |   1.04 s | 374.5 ms | 378.1 ms |   1.05 s |   1.04 s | 370.6 ms |
| 24 × 40  | 246.9 ms | 242.2 ms |   1.67 s |   1.90 s |   1.87 s |   1.71 s |   1.67 s |   1.88 s |
| 48 × 80  |   3.02 s |   3.02 s |  12.64 s |  13.48 s |  13.17 s |  12.59 s |  12.58 s |  13.18 s |

### How to read the table

**Reproducibility vs typicality.** `random.seed(42)` is set in `placement_inputs` per size, and `Input.random` → `Instance.from_components` → `pyscipopt.Model` are all deterministic, as is SCIP in sequential mode. So these numbers are **fully reproducible**: anyone running `task …:bench` on a comparable machine sees the same SCIP B&B paths and ms values (modulo OS-level noise). However, the Plant Placement problem's difficulty depends strongly on the specific random draw — plant positions, east/west balance, capacity distribution — and a quick 5-seed sweep at 12×20 shows the "slow" variants swinging by 3–5× between seeds:

| seed | (3) c+δ+card | (4) δ | (5) δ+card | (6) both | (7) both+card | (8) bigm |
|---|---|---|---|---|---|---|
| **42 (table above)** | 501 | 654 | 721 | 603 | 499 | 729 ms |
| 1    | 132 | 178 | 182 | 134 | 135 | 178 ms |
| 7    | 143 | 148 | 139 | 162 | 143 | 139 ms |
| 100  | 151 | 179 | 173 | 153 | 150 | 173 ms |
| 2026 | 152 | 177 | 173 | 162 | 153 | 167 ms |

So **individual ms values are reproducible but not representative** — seed=42 happens to draw 12×20 instances that are notably hard. What we read off the main table should be the tier structure, not the fine-grained ordering within a tier.

### Observations

- **(1) and (2) are robustly the fastest, and they are tied at every size.** Adding `δ + big-M` to a SOS1-on-`c` model (going from (1) to (2)) does *not* slow SCIP down — even when `δ` has no role beyond the big-M link. The cost of `δ` per se is essentially zero.

- **The gap opens up as size grows.** At 6×10 all eight are within ~3×. By 48×80, (1)/(2) at 3s are roughly **4× faster** than the cluster at 12–13s. The SOS1-on-`c`-alone advantage is the benchmark's strongest and most robust signal, and it holds across every seed in the sweep above.

- **Any extra per-region structure layered on top of SOS1-on-`c` breaks the advantage.** The moment cardinality (3) or a second SOS1 on `δ` ((6), (7)) is added, the runtime jumps by roughly an order of magnitude compared to (1)/(2). The redundancy seems to confuse rather than help SCIP's branching. This is robust across seeds.

- **Without SOS1 on `c`, the model lands in the same slow tier.** (4), (5), (8) cluster together (within noise of each other at each size). Replacing the cardinality bound with SOS1 on `δ` is a wash.

- **Fine-grained ordering within the slow tier is instance-specific.** At 12×20 seed=42, (3)/(6)/(7) are notably slower than (4)/(5)/(8); at 24×40 seed=42, they're comparable; at 48×80 seed=42, they're slightly faster; at other 12×20 seeds, the ordering reverses or collapses. These reversals are not informative — read the table for tier-level structure, not for individual cell ordering.

Bottom line: for this problem on SCIP, the lever is **declaring SOS1 on the continuous capacities alone**, with no other per-region structure layered on top. `δ + big-M` may be present (it's a no-op in this regime), but adding either the linear cardinality bound or a second SOS1 on `δ` costs roughly an order of magnitude. Without SOS1 on `c` at all, SCIP lands in the same slower regime regardless of how cardinality is expressed.

## Test plan

- [x] `task python:ommx-pyscipopt-adapter:test` passes (40 unit tests incl. eight-way equivalence test)
- [x] `task python:ommx:test` passes — existing `from ommx.testing import ...` consumers (`highs`, `python-mip`, `pyscipopt` adapters) all green after the package-ification
- [x] `task python:ommx-pyscipopt-adapter:pyright` / `task …:lint` clean
- [x] `task python:ommx-pyscipopt-adapter:bench` runs end-to-end under `--codspeed` for all eight targets
- [x] Sphinx docs build; `autoapi/ommx/testing/placement/index.html` is generated with KaTeX-rendered math (alignment blocks wrapped in `\begin{aligned}...\end{aligned}`)


termoshtt:
@codex review

konn (approved):
LGTM!

PR #810, opened 2026-04-20
Convert SOS1 constraints to regular constraints via Big-M

termoshtt:
## Summary

- Adds `Instance::convert_sos1_to_constraints(id)` / `convert_all_sos1_to_constraints()` in Rust and Python, following the pattern of #805 (one-hot conversion). Unlike one-hot, SOS1 needs Big-M since the rewrite is not a mathematical identity, so the return type is `Vec<ConstraintID>` per SOS1 (and a `BTreeMap<Sos1ConstraintID, Vec<ConstraintID>>` for the bulk variant).
- Per variable: if `x_i` is binary with bound `[0, 1]`, reuse it as its own indicator; otherwise introduce a fresh binary `y_i` and emit `x_i - u_i y_i <= 0` and `l_i y_i - x_i <= 0` (trivial sides are skipped). One cardinality constraint `sum_i y_i - 1 <= 0` is always added. Non-binary variables with non-finite bounds or bounds that exclude 0 are rejected; validation happens before any mutation so failures leave the instance unchanged.
- Python surface mirrors the one-hot set: `RemovedSos1Constraint`, `Instance.removed_sos1_constraints`, and `sos1_constraints_df` / `removed_sos1_constraints_df` accessors. The original SOS1 is moved to `removed_sos1_constraints` with `reason = "ommx.Instance.convert_sos1_to_constraints"` and a comma-separated `constraint_ids` parameter.

## Test plan

- [x] `cargo test -p ommx --lib instance::sos1` (7 new tests: binary reuse, integer Big-M pair, trivial-side skipping, infinite-bound rejection, bound-excluding-zero rejection, missing-ID atomicity, bulk)
- [x] `cargo test -p ommx --lib` (421 tests pass, up from 414)
- [x] `cargo clippy --all-targets` clean
- [x] `task python:test` (full Python suite passes; doctests exercise single + bulk for the all-binary reuse path)
- [x] 3 additional Python integration tests: integer Big-M pair with fresh indicators, domain-excluding-zero rejection without mutation, and `removed_sos1_constraints_df` round-trip

## Notes

- Out of scope: updating solver adapters to advertise post-conversion capability. After this PR, callers can drop SOS1 support requirements by converting first.
- The `RemovedReason.reason` convention tightened in #805 is followed here (`ommx.Instance.convert_sos1_to_constraints`).


termoshtt:
@codex review

Issue #817, opened 2026-04-21
Migrate logging to tracing + OpenTelemetry for Python profiling


termoshtt:
## Motivation

Provide OTel-based profiling so that users can enable tracing, run optimization workloads in the Python SDK, and export Chrome Trace / OTLP JSON at the Jupyter-cell granularity to feed back to OMMX developers for bottleneck analysis.

This replaces the existing `log` + `pyo3-log` → Python `logging` path with `tracing` + `pyo3-tracing-opentelemetry` → Python OTel SDK.

## Design decisions

- **Rust SDK (`rust/ommx`, `rust/dataset` libs)**: emit via `tracing` only; no global subscriber.
- **Rust binaries and examples**: install `tracing_subscriber::fmt()` + `EnvFilter`; `RUST_LOG` semantics preserved.
- **Python PyO3 extension (`python/ommx`)**: bridge Rust spans into Python OTel via `pyo3-tracing-opentelemetry`. Users configure their own `TracerProvider` + exporter before the first call.
- **Breaking change**: Python `logging` forwarding via `pyo3-log` is removed. Users who want Rust logs must configure OTel (same burden level as configuring `logging` previously).
- **Span placement**: **selective / manual**. Every impl block in `python/ommx/src/` is *not* instrumented; we pick the meaningful entry points (`Artifact.pull/push`, `Instance.from_bytes`, `Instance.evaluate`, `reduce_capabilities`, solver adapter inputs/outputs, etc.). No proc macro.
- **Span naming**: rely on `tracing`'s default (bare function name). Module path is carried in `target` and surfaces as the OTel instrumentation scope. Disambiguate same-named methods via span fields (e.g. `fields(artifact_storage = ...)`) rather than custom names.
- **Service name**: `TracingBridge::new("ommx")`.

## PRs

- [x] 1. Migrate `log` → `tracing` (no behavior change; `pyo3-log` still bridges via `tracing`'s `log` feature) — #816
- [x] 2. Replace `pyo3-log` with `pyo3-tracing-opentelemetry`; add `attach_parent_context` at selected PyO3 entry points; add `#[tracing::instrument]` where the original `log::*!` calls lived to keep events inside spans; add end-to-end tests — #823
- [x] 3. Broaden span coverage — #826
- [x] 4. Jupyter cell magic for per-cell tracing (`%%ommx_trace`) — #828
- [x] 5. Script-side tracing API (`capture_trace` context manager + `@traced` decorator) — #829
[...]

PR #828, opened 2026-04-23
Add %%ommx_trace Jupyter cell magic

termoshtt:
Part of #817 (PR4).

Adds per-cell tracing for Jupyter/IPython. `%%ommx_trace` wraps a cell in an OpenTelemetry root span, collects every span emitted during the cell (including Rust spans forwarded by `pyo3-tracing-opentelemetry`), and displays a nested text tree plus a Chrome Trace Event Format JSON download link as the cell output.

## Usage

```python
%load_ext ommx.tracing

%%ommx_trace
from ommx.v1 import Instance, DecisionVariable
x = [DecisionVariable.binary(i) for i in range(5)]
instance = Instance.from_components(
    decision_variables=x,
    objective=sum(x),
    constraints={},
    sense=Instance.MINIMIZE,
)
solution = instance.evaluate({0: 1.0, 1: 0.0, 2: 1.0, 3: 0.0, 4: 1.0})
qubo, constant = instance.to_qubo()
```

Cell output (text tree, rendered live on real Rust spans):

```
└── ommx_trace_cell (11.52 ms)
    ├── evaluate (693.0 µs)
    ├── qubo_hubo_pipeline (1.0 µs)
    └── as_qubo_format (55.0 µs)
```

Plus an HTML `<a>` link with a base64-encoded `data:application/json` URL for the full Chrome Trace — opens directly in Perfetto, speedscope, or `chrome://tracing`.

## Architecture

Four small modules under `python/ommx/ommx/tracing/`:

| File | Responsibility |
| --- | --- |
| `__init__.py` | `load_ipython_extension` / `unload_ipython_extension` (public) |
| `_setup.py` | Install the `_CellSpanCollector` lazily onto the active `TracerProvider` on first traced cell. Creates an SDK provider when the default `ProxyTracerProvider` is in place; if some other non-SDK provider was already set, raises a descriptive `RuntimeError` rather than attaching to an unused local object. |
| `_collector.py` | `SpanProcessor` subclass with an active-trace gate (`begin_capture` / `end_capture`). Spans whose `trace_id` isn't currently captured are dropped immediately on `on_end`, keeping memory bounded in long-lived notebooks. |
| `_magic.py` | The `%%ommx_trace` cell magic. Starts a root span with an explicit empty OTel `Context` (fresh `trace_id`), runs the cell body through `shell.run_cell` so input transforms / top-level `await` / line magics work, re-raises on failure so automation (`nbconvert --execute`, papermill, pytest-nbval) sees traced cells fail. |
| `_render.py` | Text-tree renderer, Chrome-Trace JSON writer, and the `display(HTML(...))` glue. Filters noisy `tracing`-crate bookkeeping attributes (`busy_ns`, `idle_ns`, `code.*`), escapes the download filename for attribute context (`quote=True`), drops open spans. |

### Design notes

- **Observational, not behaviour-changing**: cell bodies run through `shell.run_cell(cell, store_history=False)` so everything that works in an untraced cell (top-level `await`, `%` line magics, input transforms) still works. Removing `%%ommx_trace` from a cell leaves its behaviour intact.
- **Fresh trace per cell**: `start_as_current_span(..., context=Context())` detaches from any ambient OTel context, so spans emitted by unrelated instrumentation never leak into the cell's rendered tree.
- **Single traceback on failure**: `run_cell` normally prints a traceback internally; if we re-raise after it, the outer `run_cell` prints it again. We temporarily point `shell.showtraceback` at a no-op during the inner call so only the outer printer fires, and IPython's traceback UI looks identical to an untraced cell's.
- **Re-raises for automation**: `result.error_before_exec or result.error_in_exec` is surfaced and re-raised after rendering, so notebook runners that rely on cell failures to stop execution behave correctly.
- **Non-destructive provider install**: the collector is `add_span_processor`'d onto the active provider. Pre-existing exporters keep receiving spans. This relies on the dynamic span-processor resolution landed in `pyo3-tracing-opentelemetry` 0.2.0.
- **Bounded state**: the collector only retains spans whose `trace_id` was opened with `begin_capture`; `end_capture` pops and clears the slot.
- **Dependencies**: `opentelemetry-sdk` and `ipython` are declared as regular (not optional) runtime deps, since the optional-extras pattern is a known footgun (users only learn they need `pip install ommx[extra]` at `%load_ext` time).

## Tests

`python/ommx-tests/tests/test_tracing_magic.py` covers:

- **Collector**: active-trace gate (outside-window spans dropped), single-use `end_capture`, thread safety under 4 concurrent workers × 20 spans.
- **Renderer**: tree-indent nesting, empty input, Chrome-Trace event shape (`ph: X`, microsecond durations, attribute roundtrip), open-span skip, XSS regression for crafted `download_filename`.
- **Setup**: `ensure_collector_installed` is idempotent; pre-existing processors continue receiving spans after the collector attaches.
- **End-to-end through an `IPython.testing.globalipapp` shell**: user code lands in the namespace, cell exceptions surface on the return tuple, the registered magic propagates the exception (`test_magic_reraises_cell_exception`), `%load_ext` actually registers the magic, ambient spans don't leak into cell traces, top-level `await` inside the cell body works.

17 new tests; full `task python:test` remains green.

## Test plan

- [x] `task python:test` green (211 core + 17 tracing)
- [x] `task format` applied; `ruff`/`pyright` clean
- [x] `book_en` and `book_ja:build` succeed (the `autoapi/ommx/tracing/index` document is referenced from `docs/api/index.rst` so Sphinx's toctree check passes)
- [x] Manual smoke test: `%%ommx_trace` on a realistic `Instance.from_components` → `evaluate` → `to_qubo` cell produces the tree above and a working download link

## Follow-up (PR5 under #817)

Documentation: OTel setup guide, `%%ommx_trace` usage walk-through, span-naming convention rule, and troubleshooting notes.

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -21,6 +21,7 @@ Utilities
    ../autoapi/ommx/testing/index
    ../autoapi/ommx/mps/index
    ../autoapi/ommx/qplib/index
+   ../autoapi/ommx/tracing/index
 
 Adapters
 --------
--- a/python/ommx-tests/tests/test_tracing_magic.py
+++ b/python/ommx-tests/tests/test_tracing_magic.py
@@ -373,3 +373,68 @@ def test_load_ipython_extension_registers_magic(ipython_shell):
     rendering."""
     ipython_shell.extension_manager.load_extension("ommx.tracing")
     assert "ommx_trace" in ipython_shell.magics_manager.magics["cell"]
+
+
+def test_run_cell_with_trace_starts_a_fresh_trace(ipython_shell, cell_collector):
+    """The cell root span must *not* inherit an ambient trace context.
+
+    If ``start_as_current_span`` is called without detaching from the
+    current context, any ambient span (from another extension,
+    instrumentation library, or leftover ``with`` block in the user
+    namespace) bleeds into the cell's ``trace_id`` and the collector
+    captures unrelated spans. This test installs an ambient span,
+    emits a sibling span under it, runs a traced cell, and asserts
+    the cell's rendered tree contains only the cell's own spans.
+    """
+    _setup.reset_for_testing()
+    try:
+        outer_tracer = trace.get_tracer("ommx-tracing-magic-test.ambient")
+        with outer_tracer.start_as_current_span("ambient_parent") as ambient:
+            ambient_trace_id = ambient.get_span_context().trace_id
+            cell_collector.begin_capture(ambient_trace_id)
+            # A sibling span on the ambient trace — must stay out of
+            # the cell's output.
+            with outer_tracer.start_as_current_span("ambient_sibling"):
+                pass
+
+            html, exc = run_cell_with_trace(
+                ipython_shell,
+                "ommx_isolation_sentinel = 1",
+            )
+
+        assert exc is None
+        assert ipython_shell.user_ns.pop("ommx_isolation_sentinel") == 1
+        # The cell got a fresh trace, so neither ``ambient_sibling``
+        # nor the still-open ``ambient_parent`` leak into the rendered
+        # tree.
+        assert "ambient_sibling" not in html
+        assert "ambient_parent" not in html
+        assert "ommx_trace_cell" in html
+        # Defensive: spans produced on the ambient trace are still
+        # captured by the ambient-trace collector slot — they didn't
+        # silently migrate into the cell's slot.
+        ambient_spans = cell_collector.end_capture(ambient_trace_id)
+        assert "ambient_sibling" in {s.name for s in ambient_spans}
+    finally:
+        _setup.reset_for_testing()
+
+
+def test_run_cell_with_trace_supports_top_level_await(ipython_shell):
+    """Using ``shell.run_cell`` (rather than ``shell.ex``) preserves the
+    full IPython cell pipeline, including top-level ``await`` — so
+    ``%%ommx_trace`` stays observational for async cells."""
+    _setup.reset_for_testing()
+    try:
+        ipython_shell.user_ns.pop("ommx_await_sentinel", None)
+        html, exc = run_cell_with_trace(
+            ipython_shell,
+            "import asyncio\n"
+            "async def _probe():\n"
+            "    return 99\n"
+            "ommx_await_sentinel = await _probe()\n",
+        )
+        assert exc is None, exc
+        assert ipython_shell.user_ns["ommx_await_sentinel"] == 99
+        assert "ommx_trace_cell" in html
+    finally:
+        _setup.reset_for_testing()
--- a/python/ommx-tests/tests/test_tracing_magic.py
+++ b/python/ommx-tests/tests/test_tracing_magic.py
@@ -330,9 +330,10 @@ def test_run_cell_with_trace_exec_user_code(ipython_shell):
 
 
 def test_run_cell_with_trace_reports_cell_exceptions(ipython_shell):
-    """A cell that raises still produces HTML output. ``run_cell`` handles
-    the traceback display; we only need to see the exception surfaced on
-    the return tuple for programmatic inspection."""
+    """A cell that raises still produces HTML output. The caller
+    (``register_magic``) is responsible for re-raising so failure
+    semantics propagate to outer automation; :func:`run_cell_with_trace`
+    itself surfaces the exception on the return tuple."""
     _setup.reset_for_testing()
     try:
         html, exc = run_cell_with_trace(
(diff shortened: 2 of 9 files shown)


termoshtt:
Thanks for the thorough reviews — addressed everything Copilot and Codex flagged in 70bf4b1:

**Critical / functionality**
- `opentelemetry-sdk` is now declared as an optional extra (`ommx[tracing]`), and the SDK import is deferred to `ensure_collector_installed` time. A missing SDK surfaces as a clear `RuntimeError` pointing at the pip extra, not a `ModuleNotFoundError` at `%load_ext`.
- Provider-replacement logic is hardened: after `set_tracer_provider`, re-read the global and raise a descriptive error if OpenTelemetry silently ignored the call (once-set-only policy). Capability check via `hasattr(provider, "add_span_processor")` instead of `isinstance`, so custom compatible providers work.
- `_CellSpanCollector` now filters on `begin_capture` / `end_capture` — spans outside the cell magic's active trace window are dropped on arrival. This closes the unbounded-memory concern for long-lived notebooks.
- Cell execution routes through `shell.run_cell(..., store_history=False)` instead of `shell.ex(...)`. IPython now handles the traceback display via its normal path (no more magic-frame clobbering), and input transforms / display hooks behave the way a user expects.

**Security**
- Download filename is escaped with `html.escape(..., quote=True)` before landing in the `download` attribute. Added a regression test that feeds in `"><script>alert(1)</script>` and asserts the tag is not produced.

**Tests / infra**
- Replaced the per-test collector with a single session-wide collector whose state is cleared between tests — avoids piling `SpanProcessor` instances on the provider.
- Added `../autoapi/ommx/tracing/index` to the Sphinx API toctree so the `test-book (ja/en)` jobs stop failing with `document isn't included in any toctree`.

All 14 magic tests pass locally; full `task python:test` green; both `book_en` and `book_ja:build` succeed.

termoshtt:
@codex review

termoshtt:
Addressed the Codex round in 20f7765:

**Re-raise traced cell exceptions (P1)**
Switched back to ``shell.ex(cell)`` so exceptions propagate naturally. The magic now renders the trace HTML first and then re-raises, so outer automation (``nbconvert --execute``, papermill, pytest-nbval) sees the failure while IPython still displays the user's traceback (preserved via ``__traceback__``). Added ``test_magic_reraises_cell_exception`` to lock this contract down.

**``end_capture`` in a ``finally`` (Codex inline)**
Moved ``end_capture`` out of the ``with tracer.start_as_current_span`` block into an outer ``finally``. The previous placement ran *before* the root span's ``on_end`` fired, which dropped the root span from the rendered tree (caught via a pre-existing test once re-raise was in play). The outer ``finally`` also guarantees the collector's active-trace set is cleared on ``KeyboardInterrupt`` / ``SystemExit``.

**Docstring drift**
Updated the ``ommx.tracing`` module docstring to say the collector is attached *lazily* on the first traced cell, matching what ``ensure_collector_installed`` actually does.

**Pyright narrowing**
Replaced the ``hasattr(provider, "add_span_processor")`` gate with ``isinstance(provider, SdkTracerProvider)`` so pyright can narrow through it. Runtime behaviour is unchanged for the cases we actually support (SDK provider or the default ``ProxyTracerProvider``).

A few comments Copilot/Codex raised against earlier commits are already resolved in the current tree:
- `opentelemetry-sdk` hard dep landed via #827 and `ipython` via 106a834d → no more optional-extra UX to defend, so `_render.py`'s top-level SDK import is fine.
- ``set_tracer_provider`` re-read + descriptive failure already in ``_setup.ensure_collector_installed``.
- ``quote=True`` on the download filename + regression test already in ``_render.render_cell_output_html``.
- Session-wide collector with state cleared between tests already in the current fixture.

Full ``task python:test`` green (209 + 15 tracing tests); both ``book_en`` and ``book_ja:build`` clean.

termoshtt:
@codex review

termoshtt:
Addressed both P2s from Codex in e28c4a5b:

**Preserve IPython cell semantics** — switched back to ``shell.run_cell(cell, store_history=False)`` so the full input pipeline (transforms, top-level ``await``, line magics) applies. ``%%ommx_trace`` is now observational: removing the magic leaves the cell behaviour unchanged. To keep the re-raise contract without double-displaying the traceback, ``shell.showtraceback`` is temporarily pointed at a no-op during the inner call so only the outer ``run_cell`` (post re-raise) prints it. Errors are pulled from ``result.error_before_exec or result.error_in_exec`` so parse-time failures propagate too. Added ``test_run_cell_with_trace_supports_top_level_await`` as a regression guard.

**Force a fresh trace for each cell** — ``start_as_current_span`` now receives ``context=Context()`` so it never inherits the ambient OTel context. Without that, an ambient span from another extension shared its trace_id with the cell root, and the collector (keyed by trace_id) pulled sibling spans into the cell's output. Added ``test_run_cell_with_trace_starts_a_fresh_trace``: it installs an ambient span, emits a sibling, runs a traced cell under the ambient ``with``, and asserts neither the ambient parent nor the sibling appear in the cell's rendered tree.

Full test suite (211 core + 17 tracing) green, pyright clean, smoke test on a real ``evaluate`` / ``to_qubo`` cell produces the expected tree.

PR #829, opened 2026-04-23
Add capture_trace + traced decorator for script-side tracing

termoshtt:
Part of #817 (PR5).

Adds the plain-Python counterpart to the `%%ommx_trace` cell magic that shipped in #828. Scripts, tests, and CI pipelines can now capture the same text tree and Chrome Trace JSON without dragging in IPython, and the cell magic itself is now implemented on top of this new core.

## Usage

**Context manager:**

```python
from ommx.tracing import capture_trace

with capture_trace() as trace:
    instance = Instance.from_bytes(blob)
    solution = instance.evaluate(state)

print(trace.text_tree())
trace.save_chrome_trace("trace.json")
```

Sample tree from a real OMMX pipeline:

```
└── ommx_trace_block (5.33 ms)
    ├── evaluate (53.0 µs)
    ├── qubo_hubo_pipeline (1.0 µs)
    └── as_qubo_format (8.0 µs)
```

**Decorator (sugar on top):**

```python
from ommx.tracing import traced

@traced(output=\"process.json\")
def process():
    instance = Instance.from_bytes(blob)
    return instance.evaluate(state)

process()
```

All three call shapes are supported: `@traced`, `@traced()`, `@traced(name=..., output=...)`. Async is auto-detected (``inspect.iscoroutinefunction``) and wrapped with ``await``-aware machinery — a plain sync wrapper would trace only coroutine creation and close the capture window before the awaited work ran.

## Public surface

| Name | Purpose |
| --- | --- |
| `capture_trace(name=\"ommx_trace_block\")` | Context manager; `__enter__` returns a `TraceResult` placeholder that `__exit__` populates. |
| `TraceResult` | `.spans: list[ReadableSpan]`, `.text_tree()`, `.chrome_trace_json()`, `.save_chrome_trace(path)`. |
| `traced(func=None, *, name=None, output=None)` | Decorator; optionally writes Chrome Trace JSON to `output` on return *and* on exception. Supports sync and `async def`. |

## Exception handling

Per #817 design: **information is never dropped**.

1. Exceptions inside the block propagate unchanged — no swallowing.
2. OTel's default `record_exception=True` + `set_status(ERROR)` on the root span decorates it with the failure info before it closes.
3. `TraceResult.spans` is populated on the exception path too, so the caller can still inspect / save the trace from an outer `try`/`except`.
4. `_render.render_text_tree` annotates failing spans with `[ERROR]` so the user can find the failing leaf at a glance:

```
└── outer (12.1 ms) [ERROR]
    └── inner (8.4 ms) [ERROR]
```

5. `@traced(output=...)` writes the trace on **both** the success path (save errors surface, so broken configuration is caught on first use) **and** the exception path (save errors are swallowed so they do not mask the user's original exception — the signal they care about most).

## Architecture

Single new module `python/ommx/ommx/tracing/_capture.py` holding `TraceResult` + `capture_trace` + `traced`, plus a small touch to `_render.py` for the `[ERROR]` marker.

**The cell magic is now implemented on top of ``capture_trace``.** ``_magic.run_cell_with_trace`` opens a ``capture_trace(\"ommx_trace_cell\")`` block and keeps only the two pieces of behaviour unique to the notebook case:

- Temporarily point ``shell.showtraceback`` at a no-op during the inner ``shell.run_cell`` so the outer ``run_cell`` prints the traceback exactly once on re-raise.
- Explicitly ``span.set_status(Status(ERROR))`` + ``record_exception`` on the root span when ``shell.run_cell`` caught the cell's exception internally — otherwise OTel would close the root with an OK status and the rendered tree would miss its ``[ERROR]`` marker even though the cell failed.

The refactor dropped ``_magic.py`` from 105 to ~120 lines (the extra lines are the ERROR-marking + docstring for why the manual set_status is needed; the fresh-context detach, ``begin_capture`` / ``end_capture``, and trace_id bookkeeping are gone).

Tracer name is unified to ``\"ommx.tracing\"`` across cell magic, context manager, and decorator — the per-entry-point distinction (``cell`` vs ``script``) was never surfaced to users anyway.

## Tests

`python/ommx-tests/tests/test_tracing_capture.py` — 19 tests:

- **Context manager**: populates result on success + on exception, exception propagates, fresh ``trace_id`` detaches from ambient spans, ``save_chrome_trace`` creates parent dirs, ``chrome_trace_json`` roundtrips.
- **Decorator**: all three call shapes, writes on success path, writes on exception path without masking the user's exception, **async function support**, save failure surfaces on success path, ``__qualname__`` default span name, ``functools.wraps`` metadata preserved.
- **Renderer**: ``[ERROR]`` marker fires for failing spans and is absent for successful ones.

`python/ommx-tests/tests/test_tracing_magic.py` picked up one extra test after the refactor:

- **Cell magic root span ERROR**: ``test_run_cell_with_trace_marks_root_span_error_on_cell_failure`` asserts the rendered tree's root line carries ``[ERROR]`` when the cell raised — the regression guard for the manual ``set_status`` path.

All 42 tracing tests (19 capture + 18 magic + 5 bridge) pass; full ``task python:test`` green; ``pyright`` clean; smoke test on a real ``Instance.from_components`` → ``evaluate`` → ``to_qubo`` pipeline produces the tree above.

## Test plan

- [x] ``task python:test`` green (230 core + 42 tracing)
- [x] ``task format`` applied; ruff + pyright clean
- [x] Manual smoke test: ``capture_trace`` and ``@traced`` (incl. async) on real OMMX workloads produce valid trees + JSON
- [x] Cell magic continues to work end-to-end after being refactored on top of ``capture_trace``

## Follow-up (PR6 under #817)

Documentation (en/ja): OTel setup guide, ``%%ommx_trace`` walk-through, ``capture_trace`` / ``@traced`` walk-through, span-naming rule, troubleshooting.

### DIFF
--- a/python/ommx-tests/tests/test_tracing_capture.py
+++ b/python/ommx-tests/tests/test_tracing_capture.py
@@ -45,19 +45,30 @@ _SESSION_COLLECTOR: _CellSpanCollector | None = None
 @pytest.fixture
 def capture_collector():
     """Attach a single collector to the session provider and clear its
-    state between tests. Mirrors the pattern used by
-    ``test_tracing_magic.cell_collector``; the two test files can
-    reuse the same underlying collector via the module-global since
-    each test does its own state reset."""
+    state between tests.
+
+    Points ``_setup._COLLECTOR`` at the shared instance so
+    ``ensure_collector_installed()`` returns it unchanged rather than
+    creating a fresh ``_CellSpanCollector`` + attaching a new
+    ``SpanProcessor`` to the provider each test. Without that,
+    processors would pile up across the suite.
+    """
     global _SESSION_COLLECTOR
-    _setup.reset_for_testing()
     if _SESSION_COLLECTOR is None:
         _SESSION_COLLECTOR = _CellSpanCollector()
         get_test_provider().add_span_processor(_SESSION_COLLECTOR)
-    _SESSION_COLLECTOR.shutdown()
-    yield _SESSION_COLLECTOR
-    _setup.reset_for_testing()
-    _SESSION_COLLECTOR.shutdown()
+    # Reuse the shared collector for any ``capture_trace`` /
+    # ``run_cell_with_trace`` call made during the test.
+    _setup._COLLECTOR = _SESSION_COLLECTOR
+    _SESSION_COLLECTOR.shutdown()  # drop state from previous test
+    try:
+        yield _SESSION_COLLECTOR
+    finally:
+        _SESSION_COLLECTOR.shutdown()
+        # Hand the cache back to ``None`` so tests that explicitly
+        # exercise ``ensure_collector_installed`` start from a clean
+        # slate.
+        _setup._COLLECTOR = None
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


termoshtt:
Good catch — PR description mentioned refactoring the cell magic on top of ``capture_trace`` as a follow-up, but it was straightforward enough to land here. Pushed ba17a6d:

- ``run_cell_with_trace`` is now a thin wrapper around ``capture_trace(\"ommx_trace_cell\")``. The fresh-context detach, ``begin_capture`` / ``end_capture`` pairing, and trace_id bookkeeping all live in ``_capture.py``; ``_magic.py`` only keeps the two cell-magic-specific bits it actually needs (``showtraceback`` suppression, re-raise on failure).
- ``_magic.py`` drops from 105 → 74 lines (``-79 / +48``).
- Unified the tracer name to ``\"ommx.tracing\"`` for both entry points; the per-entry-point distinction (``cell`` vs ``script``) was never surfaced to users anyway — the span *name* passed to ``start_as_current_span`` is what shows up in the tree.

All 33 tracing tests (17 magic + 16 capture) pass; pyright clean; manual smoke test on an ``evaluate`` / ``to_qubo`` cell produces the same HTML output.

termoshtt:
Addressed all four in a531e34:

**[P1] Async callables in ``@traced``** — Added an ``inspect.iscoroutinefunction`` branch that returns a real ``async def`` wrapper, ``await``ing the user coroutine inside the ``capture_trace`` block. A plain sync wrapper would have traced only coroutine creation and then closed the capture window before the awaited work ran, silently dropping every span. Guard: ``test_traced_decorator_supports_async_functions``.

**[P2] Preserve the user's exception when ``save_chrome_trace`` fails** — The previous ``finally``-based save could replace the user's original exception with an I/O hiccup (bad path, permission denied). Restructured to ``try/except/else``:
- **Exception path**: ``_save_best_effort`` — save errors are caught and swallowed so ``raise`` propagates the user's exception unchanged.
- **Success path**: ``_save_if_configured`` — save errors still surface so a broken configuration is noticed the first time it trips, rather than silently failing to produce traces.

Covered by ``test_traced_decorator_save_failure_does_not_mask_user_exception`` and ``test_traced_decorator_save_failure_surfaces_on_success_path``.

**Test-fixture collector accumulation** — Both ``cell_collector`` and ``capture_collector`` used to call ``_setup.reset_for_testing`` before yielding, which wiped the module cache; every ``ensure_collector_installed`` call from within ``capture_trace`` / ``run_cell_with_trace`` then created a fresh ``_CellSpanCollector`` and attached a new ``SpanProcessor`` to the session provider. Now the fixtures point ``_setup._COLLECTOR`` at the session-wide collector so it's reused, and reset the cache only in their teardown.

19 capture tests + 17 magic tests + session suite all green; pyright clean.

termoshtt:
@codex review

termoshtt:
Both addressed in 1a0ce49:

**Cell magic root span ERROR** — When ``shell.run_cell`` catches the cell's exception internally, the ``capture_trace`` block previously exited normally and the root span closed with OK status, so the ``[ERROR]`` marker only showed on Rust-side spans (if any) and never on the root. The magic now explicitly calls ``span.set_status(Status(ERROR, ...))`` and ``record_exception(cell_exc)`` on the current (root) span before leaving the capture block. ``test_run_cell_with_trace_marks_root_span_error_on_cell_failure`` asserts the root line of the rendered tree now contains ``[ERROR]`` on cell failure.

**Fixture docstring drift** — Updated the ``test_tracing_capture`` module docstring so it describes the current session-wide collector fixture (reuse + state clear + ``_setup._COLLECTOR`` swap) rather than the earlier per-test collector variant.

37 tracing tests (18 magic + 19 capture) green; pyright clean.

termoshtt:
Addressed in 053ea0c: pass ``escaped=True`` to ``root.record_exception(cell_exc, ...)``. Because ``register_magic`` re-raises ``cell_exc`` after displaying the trace, the exception *does* propagate past the root span — setting the flag lets downstream OTel consumers (OTLP, Jaeger, Chrome Trace tooling) distinguish swallowed from propagated exceptions accurately. All 18 magic tests still green.

PR #830, opened 2026-04-23
docs(tracing): add user guide page for ommx.tracing (en/ja)

termoshtt:
## Summary

- New user guide page `user_guide/tracing.md` in both `docs/en/` and `docs/ja/`, walking through `%%ommx_trace`, `capture_trace`, and `@traced`
- Covers span naming convention, wiring up an existing OTel `TracerProvider` (OTLP / Jaeger / …), and troubleshooting common gotchas (empty trees, thread/async caveats, first-call semantics)
- Executable `{code-cell}` blocks demonstrate `capture_trace` producing the nested text tree (`ommx_trace_block` → `evaluate`) and `@traced(output=...)` writing the Chrome Trace JSON to disk
- Wired into the User Guide toctree in `docs/en/index.md` and `docs/ja/index.md`

This is the final follow-up (PR6) under #817.

## Test plan

- [x] `task python:book:test:en` — builds with `-W --keep-going`, all 42 existing + new `user_guide/tracing.md` cells execute successfully
- [x] `task python:book:test:ja` — same for Japanese
- [x] Verified the executed notebook shows the nested span tree `└── ommx_trace_block (…) / └── evaluate (…)`


termoshtt:
Thanks for the review — addressed all 10 in a2bf21fb:

- **Portability (×4)**: The `capture_trace` and `@traced` examples now build their output paths from `tempfile.gettempdir()` instead of hard-coding `/tmp/...`, so they work on Windows and in restricted sandboxes.
- **`set_tracer_provider` wording (×2)**: Rewrote the "two things to keep in mind" bullets to name what actually locks in on first use — `ommx.tracing` itself calls `set_tracer_provider(SdkTracerProvider())` on the first `capture_trace` / `%%ommx_trace` if nothing is set, and the Rust → Python tracing bridge is initialized on the first instrumented Rust call — instead of attributing it to OTel's "first call wins" alone.
- **Provider-swap tone (×2)**: Dropped the "changing providers mid-program is safe" line from the First-call Semantics section. It now lists the two things that lock in (the default SDK provider install + the pyo3 subscriber cache) and recommends mutating the existing SDK provider (`add_span_processor`) instead of replacing it.
- **`@traced(output=)` exception path (×2)**: Clarified that saving on exception is **best-effort** — the decorator attempts the write but swallows any I/O failures so they don't mask the user's original exception.

docs/en/user_guide/tracing.md
```diff
@@ -0,0 +1,202 @@
...
+    jupytext_version: 1.19.1
+kernelspec:
+  display_name: ommx-update-books (3.9.23)
+  language: python
+  name: python3
+---
+
+# Tracing and Profiling
+
+OMMX emits [OpenTelemetry](https://opentelemetry.io/) spans from the Rust core (via `tracing` + `pyo3-tracing-opentelemetry`) and from selected Python entry points. Two thin wrappers in `ommx.tracing` turn that stream into something you can actually read:
```
termoshtt: これはRust側の実装詳細なので先頭に書かなくてもいいでしょう。単にOMMXはOTelのSpanを出力するといことがわかればそれでいいはずです

docs/en/user_guide/tracing.md
```diff
@@ -0,0 +1,202 @@
...
+instance = Instance.from_components(
+    decision_variables=[x, y],
+    objective=x + y,
+    constraints={},
+    sense=Instance.MAXIMIZE,
+)
+solution = instance.evaluate({0: 1.0, 1: 1.0})
+```
+
+The cell output shows two things:
```
termoshtt, line 50: ここサンプル出力を text 扱いで入れておきましょう

termoshtt: あとここのUIのフォーマットは今後改善予定で変わっていく前提ということをかいておきましょう

termoshtt:
Ran a codex review over the docs against the actual `ommx.tracing` implementation (both the Python module and the PyO3 / Rust bridge). It surfaced four factual errors, all fixed in 180d536b:

1. **Wrong `code.namespace` attribute claim**
   The "Span Naming" section claimed that the fully-qualified module path lands in a `code.namespace` span attribute in addition to the instrumentation scope. The pyo3-tracing-opentelemetry bridge does not set a `code.namespace` attribute — it only uses the scope name (`TracingBridge::new("ommx")`). Dropped the attribute claim.

2. **`SampleSet.evaluate` does not exist**
   The Span Naming section used `Instance.evaluate` vs `SampleSet.evaluate` to illustrate same-name-different-type disambiguation, but `SampleSet` has no `evaluate` method (the relevant call is `Instance.evaluate_samples`). Replaced the example with `Artifact::push` on the `OciArchive` vs `Remote` storage backends — a real case where `tracing` uses `fields(artifact_storage = "...")` to disambiguate.

3. **`Instance.from_bytes` is not instrumented**
   `Instance.from_bytes` was listed in both Span Naming and Troubleshooting as an instrumented entry point, but `rust/ommx/src/instance/serialize.rs` has no `#[tracing::instrument]` on `Instance::from_bytes`. Replaced the examples with entry points that are actually instrumented: `evaluate`, `evaluate_samples`, `reduce_capabilities`, the `Artifact` push/pull/load/save entry points, and adapter `solve`.

4. **Subscriber-caching attribution**
   The First-call Semantics section said "the pyo3 extension caches the tracing subscriber on the first instrumented Rust call" as if it were code in this repo. That behaviour actually lives in the upstream `pyo3-tracing-opentelemetry` crate. Reworded to "the underlying `pyo3-tracing-opentelemetry` bridge installs its `tracing` subscriber process-wide on the first instrumented Rust call".

Docs still build successfully with `-W --keep-going` on both `en` and `ja`.

PR #831, opened 2026-04-23
Expose Function.evaluate_bound to Python

termoshtt:
## Summary

- Expose `Function::evaluate_bound` from the Rust SDK as a Python method on `ommx.v1.Function`.
- Signature: `Function.evaluate_bound(bounds: dict[int, Bound]) -> Bound`. Missing variable IDs are treated as `Bound.unbounded()`, matching Rust semantics.
- Document the "sound over-approximation, not tight" behavior in both Rust and Python docstrings, with a concrete example (`f = x^2 - x`, `x in [0, 1]`: true range `[-1/4, 0]`, returned `[-1, 1]`) and regression tests in both languages.
- Adopt `katexit` for math rendering in the Rust docstring (matching the convention already used in `instance/indicator.rs`, `instance/penalty.rs`, etc.).

## Motivation

The Rust SDK already has `Function::evaluate_bound` and uses it internally for slack conversion and indicator constraint handling. Python callers who want to reason about a function's interval bound (e.g. deriving constraint bounds from variable bounds without materializing samples) previously had to reimplement the logic or drop down to Rust.

## Implementation notes

- PyO3 wrapper converts `dict[int, Bound]` → `ommx::Bounds` (`BTreeMap<VariableID, ommx::Bound>`) and delegates to the existing Rust method.
- Stub (`_ommx_rust/__init__.pyi`) and API reference JSON (`docs/api/api_reference.json`) regenerated via `task python:stubgen`.
- No new Rust code — this is a pure PyO3 binding addition plus documentation polish.

## Tightness caveat

The bound is computed monomial-wise and summed, so it is a **sound over-approximation** of the true range but **not guaranteed to be tight** when multiple terms share variables (the classic dependency problem in interval arithmetic). Mathematical correctness of the caveat (including the worked example) was verified via an independent review.

## Housekeeping

- Added `.claude/rules/rust-doc-math.md` documenting the repo-wide `#[cfg_attr(doc, katexit::katexit)]` convention so future rustdoc edits involving math pick it up automatically.
- Added English/Japanese release note entries under the `Unreleased` section of v3.0.x.

## Test plan

- [x] `task format`
- [x] `task python:lint`
- [x] `task python:ommx:pyright`
- [x] `task python:ommx:test` (37 passed)
- [x] `cargo test -p ommx --lib function::evaluate_bound` (5 passed, including the new non-tightness pin)
- [x] `cargo doc -p ommx --no-deps` clean (katexit annotation)
- [x] Smoke test in Python REPL against constant / linear / squared / missing-id / non-tight cases


termoshtt:
@codex review

PR #832, opened 2026-04-23
Rust SDK v3: unify error handling around ommx::Error

termoshtt:
## Summary

Unifies the error-handling surface of the `ommx` Rust crate for v3.
Public APIs converge on a single `ommx::Result<T>` where `ommx::Error` is a
re-export of `anyhow::Error`. A small set of **signal types** remain `pub`
so callers who need a discriminant can `err.downcast_ref::<T>()`.
Diagnostic detail (breadcrumbs, field context) moves out of typed enum
variants into `tracing` events emitted alongside each bail, via
consolidated `bail!` / `error!` / `ensure!` macros that fuse a
`tracing::error!` with an `anyhow::Error` built from the same format
string.

## Design rationale

The previous error surface was a mix of `anyhow::Result`, a dozen
domain-specific `thiserror` enums, and hand-rolled traceback structures
(`ParseError::context`). An empirical survey of the two largest
downstream consumers found that neither matches on any `ommx` error
discriminant — every fail path is either propagated with `?` or
flattened via `.to_string()`. Typed enums therefore impose a
maintenance tax (new variants are breaking changes,
`#[non_exhaustive]` everywhere) without a matching benefit.

The policy:

- `ommx::Error` is a re-export of `anyhow::Error`, so `ommx::Result<T>`
  and `anyhow::Result<T>` are the same type; prefer the `ommx::` spelling
  in new public signatures for brand consistency, but don't mechanically
  rewrite pre-existing `anyhow::Result<_>` sites. (An early iteration
  wrapped `ommx::Error` in a newtype, but coherence rules forced a
  `from_anyhow` bridge at every internal-to-public boundary and every
  accessor delegated straight through — the newtype was dropped in
  favour of the re-export.)
- At every fail site, use `crate::bail!` / `crate::error!` /
  `crate::ensure!` rather than the raw `anyhow::bail!` / `anyhow::anyhow!`.
  The crate macros fuse a `tracing::error!` event with the produced
  `anyhow::Error` from one format string, so observability is uniform.
  The macros render their format arguments exactly once, so side-effectful
  or move-only expressions behave the way they read.
- **Signal types** stay `pub` and downcastable but never appear in a
  function signature where avoidable: `InfeasibleDetected`,
  `DuplicatedSampleIDError`, `CoefficientError`, `BoundError`,
  `AtolError`, plus the domain-specific enums `DecisionVariableError`,
  `SubstitutionError`, `SolutionError`, `SampleSetError` (the last four
  are consumed by in-crate tests via pattern matching, so they stay
  typed).
- **Pure-diagnostic enums** — `InstanceError`, `MpsParseError`,
  `MpsWriteError`, `ParseErrorReason`, `StateValidationError`,
  `LogEncodingError`, and the former `UnknownSampleIDError` — are gone.
  Fail sites use `crate::bail!` / `crate::error!` with an `{ field =
  value, … }` block for structured tracing fields; span context from
  `#[tracing::instrument]` replaces hand-rolled breadcrumbs.
- `Option<T>` replaces `Result<T, UnknownSampleIDError>` for key-lookup
  methods, matching the actual semantics.

## Structured error types kept as narrow-domain exceptions

A small number of structured error types remain because they carry
*positional metadata* that downstream code can consume programmatically
— not just "an enum variant for this failure mode". The canonical pattern
is: the type stores position / path info, implements `std::error::Error`,
and emits its `tracing::error!` event once at the conversion boundary
(not at every caught-and-rethrown layer). It flows into `crate::Result<T>`
via `?` without friction.

- **`ommx::ParseError`** — proto-tree breadcrumbs via
  `Vec<ParseContext>`. The `Parse` trait signature stays
  `Result<_, ParseError>`. The 33 inline insta snapshots attached to the
  Traceback format also function as documentation of "what a parse error
  from this proto looks like" — rewriting them to anyhow's linear chain
  format would strip information rather than add it.
- **`ommx::qplib::QplibParseError`** — 1-based `line_num` + rendered `message`
  for any in-file parse failure. Callers can
  `err.downcast_ref::<ommx::qplib::QplibParseError>()` to produce
  editor-style diagnostics. Its
  constructor is the single place that emits `tracing::error!(line_num,
  message, …)`, so the `FromStr` impls for `ObjSense` / `VarType` /
  `ProblemType` deliberately use plain `anyhow::bail!` rather than
  `crate::bail!` to avoid double-logging when wrapped.

The permanent `RawParseError::InvalidInstance(String)` variant added
alongside the `InstanceError` removal is the canonical way to surface
instance-invariant violations during parsing, and replaces what used to
flow through `RawParseError::InstanceError`.

## Fail-site macros

```rust
// Plain message — the format string is shared between the tracing
// event and the error's Display output.
crate::bail!("invalid OBJSENSE: {s}");

// Structured tracing fields via `{ field = value, … }`. `?foo` /
// `%foo` prefixes forward to tracing's Debug / Display capture.
crate::bail!(
    { section, size },
    "invalid field size ({size}) in section '{section}'",
);

// Signal expression — no tracing event, since callers typically
// recover by downcast.
crate::bail!(InfeasibleDetected);
```

## Docs and future-proofing

- `RUST_SDK_MIGRATION_GUIDE.md` section 7 covers the new surface: the
  signal-type catalogue, the narrow-domain structured-error pattern
  (`ParseError` + `QplibParseError`), the fail-site macros, and the
  deleted-enum list with the checklist bullets to migrate from each.
- The crate-level doc in `rust/ommx/src/lib.rs` gets an "Error handling"
  section linking the signal types.
- `.claude/rules/rust-error-handling.md` codifies the policy as a scoped
  development rule (auto-loads for `rust/ommx/**/*.rs`).

## Test plan

- [x] `cargo check --workspace --all-targets`
- [x] `cargo test -p ommx --lib` — 459 pass (up from 449 on main; 10 new
      tests cover the macro single-evaluation guarantees and the
      `QplibParseError` downcast contract for every in-file failure
      path)
- [x] `cargo test -p ommx --doc` — 43 pass, 4 ignored
- [x] `cargo clippy --workspace --all-targets` — clean
- [x] `task rust:test`
- [x] `task python:test` — all suites pass
- [x] `task format` — no diffs
- [ ] Manual build of the most active downstream Rust consumers
      against this branch via `[patch.crates-io] ommx = { path = ... }`


termoshtt:
@codex review

termoshtt:
@codex review

PR #871, opened 2026-05-11
Drop legacy disk OCI dir cache from v3 import / save / load paths

termoshtt:
## Summary

Implements §12.3 **Step C** from `ARTIFACT_V3.md` and lays out the
remaining ocipkg-removal roadmap in §12.4. After this PR, the v3
SQLite Local Registry no longer maintains the v2-era on-disk OCI
Image Layout cache — SQLite + `FileBlobStore` are the sole
post-import home of every v3 artifact.

### Step C (§12.3) — drop the legacy disk OCI dir cache

* `import::archive::import_oci_archive` and `import::remote::pull_image`
  stage into a `tempfile::TempDir` under the registry root and drop
  it on return. No more promotion to
  `registry.root().join(image_name)`.
* `pull_image` short-circuits the network fetch when SQLite already
  resolves the image ref **and** the manifest blob is present in
  `FileBlobStore`. A SQLite ref with a missing blob (registry
  corruption, interrupted import) falls through to a fresh pull with
  a `tracing::warn!` rather than handing back a stale `Unchanged`
  that would later surface as an opaque `get_blob` failure.
* New `LocalArtifact::save(output)` streams `.ommx` archives directly
  from SQLite via `OciArchiveBuilder::add_blob`, asserting each
  rehash matches the recorded manifest digest.
* CLI `Push` / `Inspect` / `Save` route the `Local(name)` branch
  through `LocalArtifact::open`, which surfaces the
  `ommx artifact import` migration hint on SQLite miss. Two helpers
  share that hint: `bail_not_found_locally` always bails (used by
  `Push`); `migration_hint_if_legacy_only` returns `Ok(())` when no
  legacy dir exists, letting `Inspect` keep its remote fallback for
  genuinely-remote refs while still bailing locally for pre-v3
  artifacts.
* Python `ArtifactInner` collapses to `{Archive, Local}` (the v2
  `OciDir` reader variant is gone). `Artifact.load_archive(dir_path)`
  imports the OCI Image Layout into the user's default SQLite Local
  Registry and returns a `Local` handle; the docstring now spells
  out that side effect and points users at the archive path
  alternative for transient reads. The file case is unchanged.
  `BuilderInner::Dir` renamed to `Local` (it always wrapped
  `LocalArtifactBuilder`).

`Artifact<OciArchive>::push` is left on the legacy ocipkg path; the
unification with the new transport is **Step D** in §12.4 below.

### §12.4 — ocipkg removal milestone series (D → F → H)

The remaining ocipkg surface is broken into three independent
PR-sized steps:

* **D** — extend the Step B `oci-client` transport to
  `Artifact<OciArchive>::push` and replace `Artifact<Remote>::pull_to`
  with a native streaming pull straight to `FileBlobStore`. Drops
  ocipkg `Remote` / `RemoteBuilder` and `auth_from_env`.
* **F** — native tar reader / writer for `.ommx` archives, dropping
  `OciArchive` / `OciDir` / `OciArtifact` / `Image` / `ImageBuilder`.
  `Artifact<Base: Image>` collapses to a single struct.
* **H** — replace `ocipkg::ImageName` in public APIs with an
  `ommx::artifact::ImageRef` new type wrapping `oci_client::Reference`.
  Breaking change; must land before v3.0 stable.

D and F are independent (network vs local format); H follows both.
The doc calls out that D should not extend `FileBlobStore`'s write
API beyond Step B's surface to avoid forcing F to redesign mid-PR.

## Pre-v3 user migration

Users with only the v2 path-tree layout under `registry.root()` now
hit the `ommx artifact import` migration hint instead of either an
opaque "image not found" or a silent remote fetch. The existing
`ommx artifact import` subcommand performs the identity-preserving
migration into SQLite.

## Test plan

- [x] `cargo test -p ommx --features remote-artifact,cli`
  — 521 unit tests + 44 doctests passing
- [x] Three new Step-C invariant tests:
  - `import_oci_archive_does_not_leave_legacy_dir_behind`
  - `pull_image_short_circuits_when_ref_is_present_with_blob`
  - `pull_image_does_not_short_circuit_when_manifest_blob_is_missing`
- [x] `cargo build --bin ommx --features remote-artifact,cli,built`
- [x] `cargo fmt --check` clean; `cargo clippy` no new warnings on
  changed files
- [x] `task python:test` (full Python suite + adapters) passing
- [x] `cargo test -p ommx --features remote-artifact,built,cli --test auth_e2e`
  remains green (no semantic change to auth flow)
- [ ] Smoke: `ommx load <archive>` → `ommx inspect <ref>` →
  `ommx save <ref> <out>` round-trip on a fresh tempdir registry


termoshtt (commented):
## P1

**`pull_image` short-circuit does not validate blob presence (corruption gap)**
`remote.rs:70-74`: when `resolve_image_name` returns `Some(manifest_digest)`, the function immediately returns `RefUpdate::Unchanged` without checking that the manifest blob (or any layer blob) is actually present in `FileBlobStore`. A registry that has the SQLite ref but missing blobs — e.g., manual blob-store deletion, partial import interrupted before blob flush — silently hands back `Unchanged`. The caller then proceeds to `LocalArtifact::open_in_registry`, hits a missing blob on first read, and fails with a generic `get_blob` error rather than a "corrupt registry, try re-pulling" message. The old `legacy_path.exists()` guard had the same blind spot, but the new path is the sole guard. A `registry.blobs().exists(&manifest_digest)` probe here, with a tracing warn + fall-through to the pull on miss, would close it.

---

## P2

**`ommx inspect <legacy-only-ref>` now triggers a network call instead of bailing locally**
`bin/ommx.rs:189-191` + `bin/ommx.rs:97-128`: `ImageNameOrPath::parse` routes any ref not in SQLite to `Remote`, so `ommx inspect ghcr.io/jij-inc/ommx/my-model:v1` with the artifact only in the legacy OCI dir silently attempts a remote fetch. `bail_not_found_locally` is only wired into the `Push` handler (`bin/ommx.rs:210`). `Inspect` (`bin/ommx.rs:189`) has no legacy-dir check; the network error message gives no migration hint. At minimum `get_manifest()` in the `Remote` arm should call `bail_not_found_locally` on a registry/network error, or `Inspect` should mirror the `Push` arm's explicit match.

**`load_archive(dir_path)` silently writes to the default SQLite registry**
`python/ommx/src/artifact.rs:130-156`: when `path` is a directory, the function calls `LocalRegistry::open_default()` and `import_oci_dir`, permanently importing the dir into the user's global `~/.local/share/ommx/…` SQLite store. The method name `load_archive` implies a read. Users calling this in a loop or a test fixture will accumulate entries in their default registry. The doc-string does not mention the side effect. Either rename/deprecate this branch or add a prominent warning in the docstring.

**Concurrency: second concurrent `pull_image` may surface `Conflicted`, not `Unchanged`**
`remote.rs:67` comment claims "the second writer sees `Unchanged`", but this holds only if both pulls produce an identical manifest blob — i.e., the remote registry returns the same bytes and the ocipkg OCI Layout writes the same `index.json`. If the remote responds with a slightly different manifest (e.g., trailing whitespace, field reorder) between the two pulls, `import_oci_dir_as_ref` will compute a different `manifest_digest` and `publish_artifact_atomic` returns `Conflicted`. Under `KeepExisting`, `Conflicted` aborts and returns an error (`index.rs:451-455`). The comment should be qualified ("assuming the remote is content-stable across concurrent pulls") rather than stated as guaranteed.

**No test for tempdir cleanup or `pull_image` SQLite short-circuit**
`tests.rs` exercises `import_oci_dir_with_policy_surfaces_ref_update` for the idempotent re-import path, but there is no test that calls `pull_image` twice (mocked or otherwise) and asserts the `Unchanged` short-circuit. There is also no test asserting that `registry_root/<image_name_as_path>/` does not exist after `import_oci_archive` or `pull_image` — the key invariant of Step C.

**§12.4 D→F→H: Phase F depends on Phase D shipping a stable `FileBlobStore` API**
`ARTIFACT_V3.md §12.4`: Phase D replaces ocipkg pull with a streaming write into `FileBlobStore`; Phase F uses the same store for `Artifact<OciArchive>::push`. The doc does not state that F is blocked on D's `FileBlobStore` interface stabilising. If D lands with an internal-only `FileBlobStore` API and F needs to extend it (e.g., resumable blob upload), that dependency should be called out explicitly to avoid interface churn mid-F.

PR #891, opened 2026-05-21
Expose Instance.substitute in the Python SDK

yuyamashiro:
## Summary

Adds a `substitute` method to the Python `Instance` class, exposing the Rust `Substitute` capability through the PyO3 bindings.

Until now the Python SDK only offered `partial_evaluate` (substitute a variable with a **constant value**). There was no way to substitute a variable with an **arbitrary function expression** — which is exactly the mechanism that the built-in `log_encode` uses internally (create binary variables → build a linear expression → substitute the integer variable).

By exposing `substitute`, users can now implement their own integer-to-binary encodings (unary, one-hot, etc.) entirely from the Python SDK, instead of being limited to the default log encoding.

- New in-place method `Instance.substitute(assignments: dict[int, ToFunction])` — replaces each given decision variable with the provided function in the objective and all active constraints.
- Raises `ValueError` on cyclic / recursive assignments, or when substituting a member of an indicator / one-hot / SOS1 constraint.
- Tests demonstrating custom **unary (thermometer)** and **one-hot** encodings, plus a recursive-assignment rejection case.

### Basic: substitute a variable with an expression

```python
from ommx.v1 import Instance, DecisionVariable

x = DecisionVariable.integer(0, lower=0, upper=3, name="x")
b = [DecisionVariable.binary(i, name="b", subscripts=[i]) for i in (1, 2)]
instance = Instance.from_components(
    decision_variables=[x, *b],
    objective=x,
    constraints=[],
    sense=Instance.MAXIMIZE,
)

# Manually log-encode x0 into two binary variables, instead of calling log_encode()
instance.substitute({0: b[0] + 2 * b[1]})
instance.objective  # Function(x1 + 2*x2)
```

### Custom unary (thermometer) encoding

The decode expression is linear; validity is enforced by ordering constraints the user adds separately.

```python
x = DecisionVariable.integer(0, lower=0, upper=4, name="x")
b = [DecisionVariable.binary(i + 1, name="b", subscripts=[i]) for i in range(4)]
instance = Instance.from_components(
    decision_variables=[x, *b], objective=x, constraints={}, sense=Instance.MAXIMIZE,
)

instance.substitute({0: b[0] + b[1] + b[2] + b[3]})   # x = sum(b_i)
for i in range(3):
    instance.add_constraint(b[i + 1] - b[i] <= 0)      # b_{i+1} <= b_i
```

### Custom one-hot encoding

```python
x = DecisionVariable.integer(0, lower=0, upper=3, name="x")
b = [DecisionVariable.binary(i + 1, name="b", subscripts=[i]) for i in range(4)]
instance = Instance.from_components(
    decision_variables=[x, *b], objective=x, constraints={}, sense=Instance.MAXIMIZE,
)

instance.substitute({0: b[1] + 2 * b[2] + 3 * b[3]})   # x = sum(v * b_v)
instance.add_constraint((b[0] + b[1] + b[2] + b[3]) == 1)  # one-hot constraint
```

> Note: practical integer encodings (log / unary / one-hot) all have a **linear** decode expression and push non-linearity into constraints. `substitute` itself accepts any `Function` (including polynomials), but high-degree encodings such as Gray code expand to O(2ⁿ) terms and are not suitable as substitution targets.

## Test plan

- [x] `task python:ommx:pytest-unit` — 394 passed (incl. new `test_substitute_unary_encode`, `test_substitute_one_hot_encode`, `test_substitute_recursive_assignment_raises`)
- [x] `task python:ommx:pytest-doctest` — passed (incl. the `substitute` docstring example)
- [x] pyright / ruff lint pass; `task format` clean
- [x] Rust extension module builds; stubs regenerated via `task python:stubgen`

### DIFF
--- a/docs/api/api_reference.json
+++ b/docs/api/api_reference.json
@@ -12412,6 +12412,48 @@
               "deprecated": null
             },
             {
+              "name": "substitute",
+              "doc": "Substitute decision variables with function expressions (in-place).\n\nReplaces each given decision variable with the provided function in the\nobjective and all active constraints. This is the general substitution\nmechanism behind {meth}`~ommx.v1.Instance.log_encode`, exposed so that\nusers can implement their own integer encodings (e.g. unary, one-hot).\n\n**Args:**\n- `assignments`: A dict mapping decision variable IDs to the function\n  expressions that should replace them.\n\nRaises ``ValueError`` on cyclic or recursive assignments, or when\nsubstituting a variable that is a member of an indicator, one-hot, or\nSOS1 constraint.\n\n# Examples\n\nEncode an integer variable x0 in range $[0, 3]$ into two binary\nvariables by hand, instead of using {meth}`~ommx.v1.Instance.log_encode`:\n\n```python\n>>> from ommx.v1 import Instance, DecisionVariable\n>>> x = DecisionVariable.integer(0, lower=0, upper=3, name=\"x\")\n>>> b = [DecisionVariable.binary(i, name=\"b\", subscripts=[i]) for i in (1, 2)]\n>>> instance = Instance.from_components(\n...     decision_variables=[x, *b],\n...     objective=x,\n...     constraints=[],\n...     sense=Instance.MAXIMIZE,\n... )\n>>> instance.substitute({0: b[0] + 2 * b[1]})\n>>> instance.objective\nFunction(x1 + 2*x2)\n```",
+              "signatures": [
+                {
+                  "parameters": [
+                    {
+                      "name": "assignments",
+                      "type_": {
+                        "display": "Mapping[int, _ommx_rust.ToFunction]",
+                        "link_target": null,
+                        "children": [
+                          {
+                            "display": "int",
+                            "link_target": null,
+                            "children": []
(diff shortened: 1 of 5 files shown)


termoshtt:
Updated this PR with a fix for the error path discussed in review.

Changes:
- Changed Rust SDK in-place helpers (`ommx::substitute`, `substitute_one`, `substitute_acyclic`) to clone the current value, run the consuming `Substitute` API, and commit only on success. On error, the original value is now preserved.
- Documented that rollback-preserving in-place helpers clone internally, and that callers should use the consuming `Substitute` methods directly when they want explicit ownership/clone control.
- Added Rust regression tests for failed in-place substitution.
- Extended the Python `Instance.substitute` recursive-assignment test to assert the `Instance` remains intact after `ValueError`.

Verified:
- `cargo test -p ommx in_place_substitute`
- `cargo fmt --check`
- `uv run pytest -q python/ommx-tests/tests/test_instance.py::test_substitute_recursive_assignment_raises`
- `uv run pytest -q python/ommx-tests/tests/test_instance.py -k substitute`

PR #934, opened 2026-06-11
Add OpenSolve for manual solver_input workflows

termoshtt:
## Summary

- Add `Run.open_solve(...)` and the `OpenSolve` context manager for advanced `solver_input` workflows where users operate the backend solver model directly.
- Make `solve.decode(...)` the point where the decoded `Solution` becomes the recorded Solve output, with the Solve entry finalized when the `open_solve` context exits.
- Preserve Solve metadata for manual workflows, including reserved solve IDs, adapter options logged through `log_adapter_option(...)`, failed/interrupted attempts, diagnostics collected through context exit, OTel Solve spans, and post-close `OpenSolve.terminal_state` inspection.
- Keep detailed internal state machines for manual Solve finalization, trace handling, and diagnostics packing/recording so each terminal outcome is explicitly represented.
- Keep `Run.log_solve(...)` as the managed solve path while routing managed and manual Solve recording through Rust-side record helpers with a clearer input shape.
- Document the manual `solver_input` workflow in the English and Japanese tutorials, API reference, and release notes.
- Update generated Python stubs, API reference metadata, and experiment tests for success, failure, diagnostics, trace/span, adapter option, missing-decode behavior, and terminal-state inspection.

## Impact

Advanced users can tune or run backend solver models directly while still getting a coherent Experiment Solve record from `solve.decode(...)`. After a manual Solve scope closes, `OpenSolve.terminal_state` gives advanced callers a debug view of the final outcome and trace/diagnostics finalization state. The lower-level Rust recording API is also structured around explicit Solve record inputs, which keeps clippy clean and makes the internal recording boundary clearer.


termoshtt:
@codex review again

termoshtt:
@codex review again

PR #953, opened 2026-06-19
Restore finite non-zero coefficient arithmetic

termoshtt:
## Summary

- Restore `Coefficient` as a finite non-zero value type with fallible arithmetic that returns `Result<Option<Coefficient>, CoefficientError>`.
- Treat cancellation and underflow-to-zero as term removal while propagating NaN and infinity as `CoefficientError` instead of panicking during normal arithmetic.
- Make `PolynomialBase` and `Function` arithmetic fallible, including term insertion, scaling, multiplication, substitution, reduction, and related instance transformations.
- Normalize `Function` arithmetic results back to the lowest matching variant (`Zero`, `Constant`, `Linear`, `Quadratic`, or `Polynomial`) after polynomial-level computation.
- Update PyO3 bindings so Python expression construction and comparisons propagate coefficient arithmetic errors instead of assuming infallible Rust operators.
- Replace infallible polynomial collection with `PolynomialBase::try_from_terms`, document macro panic contracts, and update Rust docs, examples, and tests for explicit fallible expression construction.

## API Impact

- `Coefficient` rejects zero and non-finite values during deserialization and no longer implements infallible assign arithmetic.
- Polynomial and function operators now return `Result<_, CoefficientError>` where coefficient arithmetic can fail.
- Zero-valued arithmetic results are represented by removed terms at the container layer rather than by storing `Coefficient(0.0)`.


termoshtt:
@codex review again

PR #956, opened 2026-06-24
Normalize Rust SDK collection sidecars for v3 protobuf

termoshtt:
## Summary
- Prepare the Rust SDK side of the v3 special-constraint protobuf redesign while keeping protobuf schema files unchanged.
- Add/update the temporary v3 proposal to document serialization roots, the `ModelingLabel` / `Provenance` split, sampled-values placement, unknown sidecar-ID rejection, and the intended follow-up PR sequence.
- Introduce shared `ModelingLabel` / `ModelingLabelStore` storage for decision variables, named functions, and constraints, while keeping constraint provenance as a constraint-only sidecar.
- Tighten collection invariants so active/removed ID overlap, invalid removed-reason references, duplicate insertion, and orphan label/provenance IDs are rejected at collection or top-level builder boundaries.
- Propagate fallible collection insertion and validation through instance builders, special-constraint attachment paths, parse/evaluate/sample flows, and the Python special-constraint APIs.

## Notes
- This PR is the Rust SDK normalization step before the protobuf schema PR.
- Public protobuf serialization remains scoped to top-level roots such as `Instance`, `ParametricInstance`, `Solution`, and `SampleSet`; leaf values like `Function` remain schema components rather than standalone SDK serialization roots.


termoshtt:
@codex review again

PR #957, opened 2026-06-25
Normalize label and constraint-context ownership

termoshtt:
## Summary

- Rename former element metadata sidecars to modeling labels and constraint context across the Rust SDK and Python bindings.
- Keep `ModelingLabel` as the owner-checked label payload for decision variables and named functions, and use `ConstraintContext` for constraint labels plus transformation provenance.
- Route Rust builders, collection constructors, owner setters, Python `from_components`, and attached-object setters through the same owner boundary.
- Preserve `ProcessMetadata`, artifact metadata, and extension annotation metadata as separate concepts.

## Notes

This follows #956, which normalized the Rust SDK collection sidecar ownership. This PR keeps that ownership model and renames the sidecar concepts to modeling labels and constraint context.

This is also the Rust SDK normalization follow-up to the special-constraints protobuf proposal. It keeps independent constraint-family stores while making label/context ownership explicit before the proto wrapper shape is finalized.


termoshtt:
@codex review again

PR #979, opened 2026-07-01
Move Python SDK exports to top-level ommx

### LINKED ISSUE #977: Move Python SDK domain classes to top-level ommx namespace
## Background

The protobuf wire format is expected to move away from the current `ommx.v1`
artifact/media-type contract when the new top-level serialization units are
introduced. If the wire format becomes `ommx.v2`, keeping Python SDK domain
classes under `ommx.v1` (for example `ommx.v1.Instance`) becomes misleading:
the Python class namespace would look like a protobuf-package version rather
than the stable SDK domain API.

Current state on `main`:

- PyO3 classes are registered in `ommx._ommx_rust`.
- `python/ommx/src/lib.rs` uses `pyo3_stub_gen::reexport_module_members!` to
  re-export message/domain classes into `ommx.v1`.
- `python/ommx/ommx/v1/__init__.py` is generated and re-exports `Instance`,
  `ParametricInstance`, `Solution`, `SampleSet`, `Function`, `DecisionVariable`,
  constraints, evaluated/sampled rows, and related enums from `ommx._ommx_rust`.
- `python/ommx/ommx/__init__.py` currently exposes only
  `get_default_atol` / `set_default_atol`.
- Adapters, examples, docs, migration guides, and generated stubs widely import
  user-facing classes from `ommx.v1`.
- The Python SDK migration guide is currently maintained as the standalone
  repository-root Markdown file `PYTHON_SDK_MIGRATION_GUIDE.md`, and release
  notes link to that GitHub blob directly.

## Proposed direction

Move the public Python SDK domain API from `ommx.v1` to top-level `ommx`.

Examples:

```python
from ommx import Instance, DecisionVariable, Function, Constraint
from ommx import Solution, SampleSet, State, Samples
```

The `ommx.v1` / future `ommx.v2` naming should be reserved for protobuf wire
format concepts, if exposed at all, rather than for the primary Python modeling
and SDK object model.

`ommx._ommx_rust` should remain binding-internal.

Do not keep `ommx.v1` as a re-export or compatibility alias for the top-level
SDK domain classes. A versioned Python module name should not continue to expose
the current domain object model once the wire format has moved to a new major
version. If practical, keep a minimal `ommx.v1` module that immediately raises
an `ImportError` with a short migration message pointing users to the Python SDK
migration guide; otherwise remove the module entirely.

## Scope

- Update PyO3/stub generation so the main domain classes are re-exported from
  top-level `ommx`.
- Stop generating `ommx.v1` re-exports. Prefer replacing
  `python/ommx/ommx/v1/__init__.py` with a small import-time error that directs
  users to the migration guide, instead of silently aliasing old imports.
- Update adapters and tests to import from `ommx` instead of `ommx.v1`.
- Update Python SDK docs, tutorials, migration guide, release notes, and API
  reference links from `ommx.v1.*` to `ommx.*` where they refer to SDK domain
  classes rather than protobuf wire format.
- Move the Python SDK migration guide into the Sphinx documentation tree and
  use that Sphinx page as the target for the `ommx.v1` import error message.
  The repository-root standalone Markdown should either become a short redirect
  stub or be removed after all links are updated.
- Keep subpackages with distinct responsibilities, such as `ommx.artifact`,
  `ommx.experiment`, `ommx.adapter`, `ommx.testing`, and `ommx.tracing`.
- Ensure generated `.pyi` stubs and `__all__` match the new public surface.

## Open questions

- Can we keep an `ommx.v1` module that raises a clear `ImportError` without
  interfering with packaging, stub generation, and documentation builds?
- Where should the Sphinx-integrated Python migration guide live in the docs
  hierarchy, and do we need separate English/Japanese pages or one canonical
  English page for SDK migration?
- Which names, if any, should remain under versioned modules for explicit
  protobuf/wire-format operations?

## Acceptance criteria

- Users can construct and use the primary SDK objects from top-level `ommx`.
- `from ommx.v1 import Instance` no longer succeeds as a compatibility alias;
  it either raises a migration-focused `ImportError` or the module is absent.
- The Python SDK migration guide is part of the Sphinx docs, and old references
  to `PYTHON_SDK_MIGRATION_GUIDE.md` are updated or intentionally redirected.
- User-facing adapters, examples, and docs no longer teach `from ommx.v1 import ...`
  for domain classes.
- The distinction between Python SDK object namespace and protobuf wire-format
  version is documented.
- `ommx._ommx_rust` remains internal and is not used by adapters or examples.

termoshtt:
Closes #977.

This moves the public Python SDK domain API from `ommx.v1` to top-level `ommx`, keeping `_ommx_rust` as the internal PyO3 extension module. The generated top-level `ommx` package now re-exports the public Rust-backed classes, enums, and ATol helpers, while `ommx.v1` raises an import-time migration error instead of preserving compatibility aliases.

The pyo3-stub-gen setup now generates `ommx/__init__.py` as the public surface, refreshes the stubs and API reference data for top-level `ommx`, and avoids generated RST headings that Git mistakes for conflict markers. Adapters, tests, Pyodide smoke code, and Python utilities now import from top-level `ommx`.

The documentation, migration guide, API index, release notes, and tutorial/user-guide examples have been updated to describe `ommx.v1` as the protobuf wire namespace and top-level `ommx` as the Python SDK import surface.


termoshtt:
@codex review again

PR #990, opened 2026-07-03
perf: avoid Polynomial round-trips and rehashing in Function/PolynomialBase arithmetic

konn:
## Why

Profiling JijModeling's compiler (which lowers models to OMMX by folding
per-element terms into `Function`s and multiplying functions for quadratic
objectives) attributed the dominant conversion cost to `ommx`'s polynomial
arithmetic: monomial-map rehash-and-grow cycles in `PolynomialBase`
products, and — specific to the v3 code — the `Function`-level round-trip
through `Polynomial<MonomialDyn>` performed by `normalize()` on every
arithmetic operation.

### 1. `Function::normalize` without the `Polynomial` round-trip

`Add`/`Sub`/`Div` on `Function` used to call
`from_polynomial(into_polynomial(...))`, i.e. rebuild the whole term map
with `MonomialDyn` keys and then rebuild it *again* back to the target
variant — even when nothing downgrades. Each op was O(result) instead of
O(rhs), so a fold `Σᵢ fᵢ` over N small functions was quadratic in
hashing/allocation work (see benchmarks: 1.6 s to fold 10k 3-term linear
functions).

`normalize()` now inspects the keys already stored in the map (stopping at
the first monomial of maximal degree) and only rebuilds when the variant is
actually downgraded (e.g. all quadratic terms cancelled). `Add`/`Sub`/`Div`
are routed through `try_add_assign_in_place` / in-place coefficient
updates. `Sub` is implemented as `x + (-y)`, which is bit-identical to the
previous term-map subtraction (that also added `-c` per term; the existing
`sub_via_add_neg` property test pins this).

`from_polynomial` / `into_polynomial` (crate-private) lost their last
callers and are removed. The Python binding's `Function.from_polynomial`
uses `Function::from` and is unaffected.

### 2. Pre-sized / in-place `PolynomialBase` operations

* generic `Mul`: the output map is pre-sized to `n·m` (the upper bound of
  the product's term count) instead of growing through repeated
  rehash-and-grow cycles;
* `Mul<Coefficient>` / `Div<Coefficient>`: scale coefficients in place
  instead of building a second map (shared helper also used by the new
  in-place `Function` division);
* the cross-degree `&Polynomial + &Linear`-style `Add` impls are routed
  through `try_add_assign_in_place` instead of hand-rolled loops;
* new public API: `PolynomialBase::{with_capacity, reserve}`, useful for
  downstream code that accumulates terms into a polynomial (JijModeling's
  compiler wants exactly this for its conversion fast paths).

An earlier draft also reserved `rhs.num_terms()` inside
`try_add_assign_in_place`; benchmarks showed it *hurt* merges of two
similarly-sized large maps (up to +88% — the reserve forces the table to
the sum of both sizes although the union is much smaller), so the add path
is left untouched.

### 3. No `Box<dyn Iterator>` per monomial

`Monomial::ids()` becomes an RPITIT
(`fn ids(&self) -> impl Iterator<Item = VariableID> + '_`), and
`LinearMonomial::iter` / `QuadraticMonomial::iter` return concrete
iterators. These were one boxed allocation per monomial inside
`required_ids()` / `evaluate()` loops and inside every
quadratic-monomial product (visible in the `square-quadratic` results
below).

## Numerical caveat

Per-monomial results of every single operation are unchanged (the
accumulation loop order within one operation is untouched). However,
pre-sized hash maps iterate in a different order than organically grown
ones, so when an intermediate result flows into a *later* product whose
output monomial receives ≥ 3 contributions (degree-≥3 chains), the
floating-point accumulation order — hence possibly the last ulp — can
differ from the previous release, as can serialized term order.
Run-to-run determinism is unaffected (FNV is unseeded). JijModeling has
already accepted the same caveat for its own pre-sized conversion fast
paths.

## Benchmarks

`cargo bench -p ommx --bench sum` / `--bench mul` with
`--warm-up-time 1 --measurement-time 2`, aarch64 dev container (6 cores),
vs. a criterion baseline saved on `main` (d7c1325) with the same bench
additions applied.

This PR adds three Function-level bench groups
(`sum-function-linear-small-many`, `sum-function-quadratic-small-many`,
`square-function-linear`): the existing suites only exercised
`PolynomialBase` and never crossed the `normalize()` path that dominates
`Function` arithmetic.

**Function-level (new benches):**

| benchmark | main | this PR | change |
|---|---:|---:|---:|
| sum-function-linear-small-many/100 | 187.6 µs | 5.6 µs | −97.1% |
| sum-function-linear-small-many/1000 | 16.75 ms | 62.3 µs | −99.6% |
| sum-function-linear-small-many/10000 | 1.615 s | 744 µs | −99.95% (≈2200×) |
| sum-function-quadratic-small-many/100 | 452.8 µs | 12.8 µs | −97.2% |
| sum-function-quadratic-small-many/1000 | 46.09 ms | 138.9 µs | −99.7% |
| sum-function-quadratic-small-many/10000 | 5.400 s | 1.408 ms | −99.97% (≈3800×) |
| square-function-linear/10 | 4.47 µs | 1.47 µs | −67% |
| square-function-linear/100 | 463.8 µs | 201.3 µs | −56% |

(The quadratic growth of the `main` column across 100→1000→10000 is the
per-op round-trip; the patched column grows linearly.)

**PolynomialBase-level (existing benches):**

| benchmark | main | this PR | change |
|---|---:|---:|---:|
| square-linear/10 | 2.18 µs | 1.50 µs | −32% |
| square-linear/100 | 360.1 µs | 315.5 µs | −8% |
| square-quadratic/10 | 13.63 µs | 8.16 µs | −40% |
| square-quadratic/100 | 1.420 ms | 1.042 ms | −27% |
| square-polynomial/10 | 12.87 µs | 10.31 µs | −19% |
| square-polynomial/100 | 1.304 ms | 1.140 ms | −13% |

The `sum-*` / `add-*` PolynomialBase suites are unchanged within observed
run-to-run variance (±10% in this environment; the add path is
byte-identical to `main`).

## Testing

* `cargo test -p ommx` — all pass (751 unit tests + doctests), including
  the property tests for add/sub/mul ref-parity, commutativity,
  associativity, `sub_via_add_neg`, and the normalization test
  `arithmetic_normalizes_low_degree_results`.
* `cargo clippy --workspace --all-features` — no new warnings (the single
  remaining warning in `instance/penalty.rs` is pre-existing on `main`).
* `cargo fmt --all --check`, `cargo doc --no-deps -p ommx` — clean.


termoshtt (approved):
LGTM

Issue #1046, opened 2026-07-09
Add benchmarks for generic substitute helpers on Instance/ParametricInstance


termoshtt:
## Summary

Split out from #1031. Follows the same "benchmark evidence before
implementation issue" sequencing as #1044 and #1045.

`ommx::substitute`, `substitute_one`, and `substitute_acyclic` (in
`rust/ommx/src/substitute.rs`) clone the entire `T` before calling the
consuming `Substitute` API and committing the clone back on success. This is
fine for small expressions, but potentially expensive for `T = Instance` or
`T = ParametricInstance`. Python `Instance.substitute(...)` and
`ParametricInstance.substitute(...)` use these helpers, so large instances
can hit the same broad-clone pattern already benchmarked for
`reduce_binary_power` (#1044) and `log_encode` (#1045).

## Suggested first slice

Add criterion benchmarks under `rust/ommx/benches/` for the in-place
`substitute`/`substitute_one`/`substitute_acyclic` free functions applied to
`Instance` (and `ParametricInstance` if convenient), using the same shape as
`rust/ommx/benches/reduce_binary_power.rs` and
`rust/ommx/benches/log_encode.rs`: hold the real substitution work constant
(substitute one or a few variables) while growing unrelated removed/active
constraint count, to see whether cost scales linearly with unrelated state.

## Suggested direction (pending benchmark confirmation)

If the benchmarks show the same clone-dominated linear scaling, add
owner-specific private plan/commit substitution paths for `Instance` and
`ParametricInstance`, while keeping the generic helper as a conservative
fallback for arbitrary `T`. This is Instance-internal; no public Plan/Commit
API should be exposed. File that as a separate implementation issue once the
benchmark evidence is in.

## Non-goals

- Do not change #1027's `Instance::partial_evaluate` workstream.
- Do not expose a public Plan/Commit API.
- Do not claim performance improvements without benchmark results.
- Do not change public APIs until this issue's benchmark evidence justifies
  a follow-up implementation issue.

## Acceptance criteria

- Benchmark coverage exists for the clone-backed `substitute`/
  `substitute_one`/`substitute_acyclic` shape on `Instance` (and ideally
  `ParametricInstance`).
- A follow-up implementation issue is filed only if the benchmarks show
  clone-dominated scaling.

termoshtt:
Status after merged PR #1047: this issue remains open and its benchmark-first
scope is still valid.

[...]

Issue #1045, opened 2026-07-09
Give Instance::log_encode an owner-specific private plan/commit path


termoshtt:
## Summary

Split out from #1031 after benchmark evidence. Related to #1044
(`Instance::reduce_binary_power`), which has the same clone-then-commit shape
and is expected to land the shared private mechanism first.

`Instance::log_encode` clones the whole `Instance` once, encodes the
requested integer variable(s) on the clone (new binary variables + a
consuming `substitute`), then commits the clone back on success. The clone
pays for removed constraints, special constraints, named functions,
metadata, and other state unrelated to the encoded variable(s). This path is
user-facing and is also used by adapter preparation flows.

## Benchmark evidence

Added in #1031 as `rust/ommx/benches/log_encode.rs`. Both benchmarks hold the
real encoding work constant (one integer variable log-encoded into ~10 bits)
and grow only unrelated instance state:

- `log-encode-removed-constraints` (unrelated removed regular constraints):
  ~88µs at 1,000 constraints -> ~3.0ms at 32,000.
- `log-encode-active-constraints` (unrelated active regular constraints not
  referencing the encoded variable): ~125µs at 1,000 -> ~3.7ms at 32,000.

Cost scales linearly with unrelated instance size for a fixed amount of real
encoding work, consistent with the whole-instance clone dominating cost.

## Suggested direction

Give `log_encode` an instance-owned rewrite path that adds the auxiliary
binary variables and substitutes the selected variable(s) without cloning
the whole `Instance` up front. A bulk-oriented plan (encode all requested
variables in one pass) may be preferable to repeating a clone-prone
single-variable path. This is Instance-internal; no public Plan/Commit API
should be exposed.

## Non-goals

- Do not change #1027's `Instance::partial_evaluate` workstream.
- Do not expose a public Plan/Commit API.
- Do not claim performance improvements without new benchmark results
  comparing before/after.

## Acceptance criteria

- `log_encode`'s cost for a fixed amount of encoding work is no longer
  linear in unrelated removed/active constraint count, verified against the
  existing `rust/ommx/benches/log_encode.rs` benchmarks.
- Existing `log_encode` correctness tests in
  `rust/ommx/src/instance/log_encode.rs` (including evaluation-preserving
  and fixed-variable-rejection cases) continue to pass unchanged.

termoshtt:
## Flamegraph analysis (CodSpeed instrumented run)

[...]

Issue #1044, opened 2026-07-09
Give Instance::reduce_binary_power an owner-specific private plan/commit path


termoshtt:
## Summary

Split out from #1031 after benchmark evidence.

`Instance::reduce_binary_power` clones the whole `Instance`, rewrites the
objective and active regular constraints on the clone, then commits the clone
back on success. The clone pays for removed constraints, special constraints,
named functions, metadata, and other state the operation never touches.

## Benchmark evidence

Added in #1031 as `rust/ommx/benches/reduce_binary_power.rs`. Both benchmarks
hold the real rewrite work constant (a single `x0^2` binary-power term in the
objective) and grow only unrelated instance state:

- `reduce-binary-power-removed-constraints` (unrelated removed regular
  constraints): ~87µs at 1,000 constraints -> ~2.8ms at 32,000.
- `reduce-binary-power-active-constraints` (unrelated active regular
  constraints not referencing the rewritten variable): ~78µs at 1,000 ->
  ~2.7ms at 32,000.

Cost scales linearly with unrelated instance size for a fixed amount of real
rewrite work, consistent with the whole-instance clone dominating cost.

## Suggested direction

Replace the whole-instance clone with a private plan containing only the
rewritten objective and active-constraint replacements, committed back after
all coefficient-failure paths have already been checked (matching current
rollback-safety behavior). This is Instance-internal; no public Plan/Commit
API should be exposed.

## Non-goals

- Do not change #1027's `Instance::partial_evaluate` workstream.
- Do not expose a public Plan/Commit API.
- Do not claim performance improvements without new benchmark results
  comparing before/after.

## Acceptance criteria

- `reduce_binary_power`'s cost for a fixed rewrite is no longer linear in
  unrelated removed/active constraint count, verified against the existing
  `rust/ommx/benches/reduce_binary_power.rs` benchmarks.
- Rollback-on-error behavior is preserved (see
  `reduce_binary_power_preserves_instance_on_coefficient_error` in
  `rust/ommx/src/instance/reduce_binary_power.rs`).

termoshtt:
## Flamegraph analysis (CodSpeed instrumented run)

Queried the CodSpeed flamegraph for `reduce-binary-power-removed-constraints[32000]` (31.1ms total) to check where the linear-scaling cost from #1031's benchmark actually goes.

[...]

Issue #1031, opened 2026-07-09
Track clone-backed rollback overhead outside partial_evaluate


termoshtt:
## Summary

Split out from #1027 to track clone-backed rollback overhead outside
`Instance::partial_evaluate`.

The benchmark-first investigation and the measured owner-specific fixes are now
complete. Remaining work has been separated into focused issues rather than
keeping this discovery workstream open.

## Final status

| Area | Status | Tracking |
| --- | --- | --- |
| `Instance::reduce_binary_power` clone/drop overhead | Completed in #1047 | #1044 |
| `Instance::log_encode` clone/drop overhead | Completed in #1047 | #1045 |
| Generic clone-backed `substitute` helpers on large owners | Benchmark first; remains open | #1046 |
| `NamedFunctionTable::partial_evaluate` | No demonstrated workload or dedicated benchmark; revisit only with evidence | — |

PR #1047 merged as `0b66b964789b13802334a124da21e06f0faca43d`.
It replaced detached whole-owner rollback clones with private owner-level
planning and narrow commits. The final `log_encode` path also updates
`AcyclicAssignments` atomically without cloning or rewriting unrelated
dependency right-hand sides.

## Final benchmark evidence

Final-head CodSpeed workflow:
https://github.com/Jij-Inc/ommx/actions/runs/29072627767

Same-run cross-size exponents from 1,000 to 32,000 rows:

| Family | Exponent |
| --- | ---: |
| `reduce_binary_power`, removed constraints | `p = 0.001` |
| `reduce_binary_power`, active constraints | `p = 0.990` |
| `log_encode`, removed constraints | `p = 0.003` |
| `log_encode`, active constraints | `p = 0.956` |

The removed-row families are guardrails against accidental whole-owner work
and are effectively constant. The active-row families remain linear because
inspecting active rows is required; those exponents are complexity guardrails,
not evidence of constant-factor speedup.

The early and final PR runs are not valid direct before/after timing pairs:
their fixtures differ and the CodSpeed runner hardware changed. Earlier
speedup tables in child-issue comments are therefore superseded by the final
same-run cost-model evidence above.

## Remaining and related work

- #1046 remains open for benchmark coverage of the generic free
  `substitute`/`substitute_one`/`substitute_acyclic` helpers on `Instance` and
  `ParametricInstance`. PR #1047 intentionally did not change their
  clone-backed atomic contract.
- #1051 tracks a separate correctness gap: `reduce_binary_power` does not yet
[...]

PR #1047, opened 2026-07-09
Add benchmarks and fix clone overhead for reduce_binary_power and log_encode

termoshtt:
## Summary

- Adds persistent Criterion/CodSpeed guardrails for `Instance::reduce_binary_power` and `Instance::log_encode` with deterministic input shapes and explicit benchmark contracts.
- Replaces `reduce_binary_power`'s whole-`Instance` rollback clone with an owner-level plan containing only the rewritten objective and changed active regular-constraint rows. All coefficient failures occur before commit.
- Replaces `log_encode`'s whole-owner clone with a bulk plan for auxiliary variables and affected expressions. The plan and commit phases are private and can only be applied to the same `Instance` receiver.
- Updates `AcyclicAssignments` atomically without cloning unrelated dependency right-hand sides. The consuming `Instance: Substitute` path moves the dependency table by value, while borrowed log encoding clones only affected assignments and rebuilds the dependency graph from borrowed unchanged rows.
- Keeps named-function labels and constraint sidecars owned by their tables through narrow validated row-replacement effects.

## Benchmark contracts

- Removed-constraint families are scaling guardrails for accidental whole-owner work. With removed rows outside the operation boundary, the expected same-run exponent is `p ~= 0`.
- Active-constraint families are complexity guardrails for the required row scan. Their expected exponent is `p ~= 1`; this is not used as evidence for constant-factor clone removal.
- The 32,000 unrelated named-functions case and 8,000 unrelated dependency-assignments case are fixed-input regression baselines. Profiles are diagnostic only after a same-input regression is detected.
- These persistent guardrails run on `main` and release tags; full performance-PR runs remain manual.

## CodSpeed evidence

Final-head CodSpeed simulation: https://github.com/Jij-Inc/ommx/actions/runs/29072627767

| Family | Same-run evidence |
| --- | ---: |
| `reduce_binary_power`, removed constraints | `p = 0.001` |
| `reduce_binary_power`, active constraints | `p = 0.990` |
| `log_encode`, removed constraints | `p = 0.003` |
| `log_encode`, active constraints | `p = 0.956` |
| `log_encode`, 32,000 unrelated named functions | `T = 13.39 ms` |
| `log_encode`, 8,000 unrelated dependencies | `T = 17.05 ms` |

The original benchmark commit and final head are not valid before/after timing pairs: the original fixtures also grew the decision-variable count, the final fixtures keep it fixed, and the CodSpeed runner hardware changed. This PR therefore makes no direct speedup claim from those runs. The same-run exponents above validate the final cost models, and the fixed-input URIs establish comparable baselines for future commits.

## Scope

- Implements the owner-specific clone-removal work tracked in #1044 and #1045 while keeping #1031 as the parent workstream.
- Leaves generic free-function substitution benchmarking to #1046.
- Leaves full `reduce_binary_power` coverage for active indicator bodies, named functions, and dependency right-hand sides to #1051.
- Does not expose a public Plan/Commit API and does not change #1027's `partial_evaluate` workstream.


termoshtt:
## Update: this PR now also fixes the clone overhead, not just benchmarks it

Following up on the flamegraph analysis posted to #1044 and #1045 (which showed ~92-97% of the benchmarked cost at 32,000 unrelated constraints was the whole-instance `clone` + the `drop` of the old instance at commit time, not the actual rewrite), this PR now includes the fix for both methods, commit a3756e0c.

### What changed

- **`Instance::reduce_binary_power`**: instead of cloning the whole `Instance` and swapping it in via `*self = updated`, it now computes the rewritten objective and active-constraint replacements on copies of only the touched fields, and writes them directly into `self` after all fallible steps succeed. Removed constraints, special constraints, named functions, and metadata are never cloned or dropped.
- **`Instance::log_encode`**: validates every requested id against `self` up front — including the indicator/one-hot/SOS1 checks `Substitute::substitute_acyclic` would otherwise perform mid-substitution — before mutating anything. Once validation succeeds, the commit phase (creating auxiliary binary variables + substituting them) is infallible, so it can mutate `self` directly via a new private `Instance::substitute_acyclic_mut` in-place helper instead of cloning.
- Existing rollback-safety tests (including `test_log_encode_is_atomic_when_later_id_fails`, `test_log_encode_is_atomic_when_substitution_fails`, `test_log_encode_rejects_indicator_variable_without_side_effects`, and `reduce_binary_power_preserves_instance_on_coefficient_error`) all pass unchanged, confirming atomicity is preserved.
- No public API changed; `Instance::substitute_acyclic_mut` is a private, crate-internal helper (`pub(crate)`).

### Before / after (local criterion wall-clock, 32,000 unrelated removed constraints)

| Benchmark | Before | After | Speedup |
| --- | --- | --- | --- |
| `reduce-binary-power-removed-constraints[32000]` | 2.8 ms | 49 µs | ~57x |
| `reduce-binary-power-active-constraints[32000]` | 2.7 ms | 1.04 ms | ~2.6x (bounded by having to inspect every active constraint, which is inherent work) |
| `log-encode-removed-constraints[32000]` | 3.0 ms | 8.8 µs | ~340x |
| `log-encode-active-constraints[32000]` | 3.7 ms | 2.07 ms | ~1.8x (bounded by `substitute`'s need to check every active constraint for whether it's affected) |

The "active-constraints" cases improve less because inspecting every active constraint for a possible match is inherent, necessary work for both methods — only the previously-dominant clone/drop of *unrelated* state (removed constraints, special constraints, named functions, metadata) is eliminated.

Also re-triggered the CodSpeed benchmark workflow on this commit for an instrumented before/after comparison: https://github.com/Jij-Inc/ommx/actions/runs/29024027960

This closes out #1044 and #1045 pending review; will comment on both issues with a summary once CI confirms.

PR #1121, opened 2026-07-23
Restructure Lean Instance semantics and formalize special-constraint lowering

termoshtt:
Rendered README: https://github.com/Jij-Inc/ommx/tree/revise-lean/lean#readme

## Summary

- reorganize the independent Lean reference model around the mathematical structure of an OMMX `Instance`: exact-rational states and affine functions, explicit domains and bounds, and parallel regular, OneHot, SOS1, and Indicator constraint families
- define `Instance` feasibility, objective values, optimization sense, sound affine domain-box bounds, and extension operations for transformations that append decision-variable components
- introduce `Instance.Transform` with an explicit target, partial encode/decode maps, directional reduction and relaxation, objective and sense preservation, independent source/target round trips, and composition
- formalize Indicator and SOS1 Big-M lowering as partial `Instance.Transform` constructors organized into formulation semantics, `Plan` validation, target construction, and top-level transformation contracts
- recompute lowering bounds from the source `Instance` and construct targets only from validated plans; Indicator lowering keeps identity state maps, while SOS1 lowering reuses binary members and appends fresh selectors with Big-M links and an at-most-one constraint
- remove the earlier generic problem/reduction/projection layer, Farkas certificate code, structural recognition and promotion prototypes, selector-isolation machinery, and monolithic `Special` modules that were ahead of the implemented scope
- update the Lean README, umbrella imports, and focused fixtures around the resulting Instance and lowering semantics

## Scope

The Lean model remains an independent exact-rational reference semantics rather than a verification of the Rust SDK or protobuf implementation. Presolve transformations, promotion and inverse lowering, infeasibility certificates, and a versioned SDK-to-Lean `Witness` verification boundary remain future work.


lean/OMMXProof/Function/Affine.lean
```diff
@@ -0,0 +1,146 @@
...
+    (domains : Fin n → Domain) : Bound :=
+  (List.ofFn fun i =>
+      Bound.scale (expr.coeff i) (domains i).bound).foldr
+    Bound.add (Bound.point expr.constant)
+
+/-- Every state in the supplied domains evaluates inside `evaluateBound`. -/
+theorem evaluateBound_sound (expr : Affine n)
+    (domains : Fin n → Domain) {state : State n}
+    (hdomains : ∀ i, state i ∈ domains i) :
+    expr.eval state ∈ expr.evaluateBound domains := by
```
kory33-jij: [nits] I think it's more natural to split this definition into `evaluateBound` + `evaluateBoundFromDomains` (applying this diff will invalidate downstream references, so please perform a global rename)
```suggestion
/-- The minimum rational interval containing every value of an affine function
over the given decision-variable bounds. -/
@[simp]
def evaluateBound (expr : Affine n) (bounds : Fin n → Bound) : Bound :=
  (List.ofFn fun i => (bounds i).scale (expr.coeff i)).foldr
    Bound.add (Bound.point expr.constant)

/-- A rational interval containing every value of an affine function over the
given decision-variable domains.

Integer domains contribute their containing rational intervals, so
this interval is in general not a tight bound for the discrete affine image.
-/
abbrev evaluateBoundFromDomains (expr : Affine n) (domains : Fin n → Domain) : Bound :=
  evaluateBound expr (fun i => (domains i).bound)

/-- Every state in the supplied domains evaluates inside `evaluateBound`. -/
theorem evaluateBoundFromDomains_sound (expr : Affine n)
    (domains : Fin n → Domain) {state : State n}
    (hdomains : ∀ i, state i ∈ domains i) :
    expr.eval state ∈ expr.evaluateBoundFromDomains domains := by
```

lean/OMMXProof/Constraint/Indicator.lean
```diff
@@ -0,0 +1,55 @@
...
+instance (constraint : IndicatorConstraint n) (state : State n) :
+    Decidable (constraint.Holds state) := by
+  unfold Holds
+  infer_instance
+
+end IndicatorConstraint
+
+def IndicatorPredicate (trigger : Fin n) (polarity : IndicatorPolarity)
+    (body : State n → Prop) (state : State n) : Prop :=
+  polarity.Active (state trigger) → body state
```
kory33-jij: [nits, imo] maybe we should define `Holds` in terms of `IndicatorPredicate`

```suggestion
def IndicatorPredicate (trigger : Fin n) (polarity : IndicatorPolarity)
    (body : State n → Prop) (state : State n) : Prop :=
  polarity.Active (state trigger) → body state

namespace IndicatorConstraint

abbrev Holds (constraint : IndicatorConstraint n) (state : State n) : Prop :=
  IndicatorPredicate constraint.trigger constraint.polarity constraint.body.Holds state

instance (constraint : IndicatorConstraint n) (state : State n) :
    Decidable (constraint.Holds state) := by
  unfold Holds IndicatorPredicate
  infer_instance

end IndicatorConstraint
```

lean/OMMXProof/Instance/Transform.lean
```diff
@@ -0,0 +1,337 @@
+import OMMXProof.Instance
```
kory33-jij: Comment on the module discipline: AFAIK it is less common to prepare `[module name].lean` as an preliminary entrypoint into `[module name]/*.lean` files. Following this I think we should invert dependency of `Transform.lean`, `Transform/IndicatorBigM.lean` and `Transform/SOS1BigM.lean` by moving them to `Transform/Basic.lean`, `Transform/IndicatorBigM/Basic.lean` etc. (which is a common pattern in libraries like Mathlib)

lean/OMMXProof/Constraint/Indicator.lean
termoshtt: Fixed in https://github.com/Jij-Inc/ommx/pull/1121/commits/384b54f9b19899147f8291743915e9bc554f45e0

lean/OMMXProof/Function/Affine.lean
termoshtt: fixed in https://github.com/Jij-Inc/ommx/pull/1121/commits/384b54f9b19899147f8291743915e9bc554f45e0

lean/OMMXProof/Instance/Transform.lean
termoshtt: Fixed in https://github.com/Jij-Inc/ommx/pull/1121/commits/dd60b1af3f77d2fe4e7e7d4133a2d3bfcc891e49 and https://github.com/Jij-Inc/ommx/pull/1121/commits/a2420341cd0dfa5704d189bbbcca1e213c4a54ca

lean/OMMXProof/Instance/Extend.lean
```diff
@@ -0,0 +1,251 @@
...
+import Mathlib.Algebra.BigOperators.Fin
+import Mathlib.Data.Fin.Tuple.Basic
+
+/-!
+# Extending an Instance with fresh components
+
+An Instance transformation may change the dimension of its state space.  This
+module embeds an `n`-component Instance into the left block of an
+`(n + k)`-component Instance and supplies the corresponding state operations.
+The fresh right block is absent from every lifted expression and constraint.
```
kory33-jij: I couldn't understand what this sentence is supposed to convey, so maybe remove it?
```suggestion
```

lean/OMMXProof/Instance/Extend.lean
```diff
@@ -0,0 +1,251 @@
...
+theorem holds_extend (constraint : LinearConstraint n)
+    (state : State (n + k)) :
+    (extend constraint k).Holds state ↔
+      constraint.Holds (State.source state) := by
+  simpa only [State.append_source_fresh] using
+    holds_extend_append constraint (State.source state) (State.fresh state)
+
+end LinearConstraint
+
+def extendMembers (members : Finset (Fin n)) (k : Nat) :
```
kory33-jij: [imo, nits] This function should probably be named `castAddFinsetFin` since here we are not `extend`ing but just coarsening the bound. IMO `OneHotConstraint.extend` should similarly be named `castAdd` or `coarsenAdd`.

```suggestion
def castAddFinsetFin (members : Finset (Fin n)) (k : Nat) :
```

lean/OMMXProof/Instance/Extend.lean
```diff
@@ -0,0 +1,251 @@
...
+
+namespace State
+
+def append (source : State n) (fresh : State k) : State (n + k) :=
+  Fin.append source fresh
+
+def source (state : State (n + k)) : State n :=
+  fun i => state (Fin.castAdd k i)
+
+def fresh (state : State (n + k)) : State k :=
```
kory33-jij: [imo, nits] There's nothing intrinsically *fresh* about a tail of a `State`, and I think this is a misnomer following how `Affine.extend` etc set `0` coefficient on the extended part. Maybe:

```suggestion
def extendedPart (state : State (n + k)) : State k :=
```

termoshtt: https://github.com/Jij-Inc/ommx/pull/1121/commits/0f164cabe540cf28a8a1f19516d45fe97cbb3e3c fixed

lean/OMMXProof/Instance/Extend.lean
termoshtt: https://github.com/Jij-Inc/ommx/pull/1121/commits/0f164cabe540cf28a8a1f19516d45fe97cbb3e3c fixed

lean/OMMXProof/Instance/Extend.lean
termoshtt: I clarify the original intent in https://github.com/Jij-Inc/ommx/pull/1121/commits/0613ec49f6447b13170b730b324988d40e143650.

lean/OMMXProof/Instance/Transform/Basic.lean
```diff
@@ -0,0 +1,337 @@
...
+
+This is the feasibility-soundness condition required of a reduction. It rules
+out transformed feasible solutions which have no meaning for the source
+Instance. -/
+def IsReduction {source : Instance n} (transform : Transform source) : Prop :=
+  ∀ {targetState},
+    transform.target.Feasible targetState →
+      ∃ sourceState,
+        transform.decode targetState = some sourceState ∧
+          source.Feasible sourceState
```
kory33-jij: [nits, style] equivalent, but I find it more idiomatic to write this as:

```suggestion
      ∃ sourceState ∈ transform.decode targetState, source.Feasible sourceState
```

Downstream proofs will break but I'm sure Claude can autonomously fix them.

lean/OMMXProof/Instance/Transform/Basic.lean
```diff
@@ -0,0 +1,337 @@
...
+/-- Every feasible source state can be encoded as a feasible target state.
+
+For identical state spaces and the identity encoding, this is the usual
+feasible-region inclusion defining a relaxation. -/
+def IsRelaxation {source : Instance n} (transform : Transform source) : Prop :=
+  ∀ {sourceState},
+    source.Feasible sourceState →
+      ∃ targetState,
+        transform.encode sourceState = some targetState ∧
+          transform.target.Feasible targetState
```
kory33-jij: [nits, style] similar to the comment above
```suggestion
      ∃ targetState ∈ transform.encode sourceState, transform.target.Feasible targetState
```

lean/OMMXProof/Instance/Transform/SOS1BigM/Formulation.lean
```diff
@@ -0,0 +1,225 @@
...
+
+def plannedSelector [DecidableEq ι] (reused : Finset ι)
+    (members freshSelectors : ι → Rat) : ι → Rat :=
+  fun i => if i ∈ reused then members i else freshSelectors i
+
+def OptionalUpperLink (upper member selector : Rat) : Prop :=
+  if 0 < upper then member ≤ upper * selector else True
+
+def OptionalLowerLink (lower member selector : Rat) : Prop :=
+  if lower < 0 then lower * selector ≤ member else True
```
kory33-jij: [imo] I think `if ... else True` should be replaced with plain implications:

```suggestion
def OptionalUpperLink (upper member selector : Rat) : Prop :=
  (0 < upper) → (member ≤ upper * selector)

def OptionalLowerLink (lower member selector : Rat) : Prop :=
  (lower < 0) → (lower * selector ≤ member)
```

lean/OMMXProof/Instance/Transform/SOS1BigM/Formulation.lean
```diff
@@ -0,0 +1,225 @@
...
+
+/-- Exact formulation of a mixed reused/fresh selector layout. -/
+def PlannedSelectorFormulation [Fintype ι] [DecidableEq ι]
+    (reused : Finset ι) (bounds : SelectorBounds ι)
+    (members freshSelectors : ι → Rat) : Prop :=
+  GenericBinaryOn Finset.univ (plannedSelector reused members freshSelectors) ∧
+    (∀ i, i ∉ reused →
+      OptionalUpperLink (bounds.upper i) (members i) (freshSelectors i) ∧
+        OptionalLowerLink (bounds.lower i) (members i) (freshSelectors i)) ∧
+    ∑ i, plannedSelector reused members freshSelectors i ≤ 1
```
kory33-jij: [imo] I think it makes sense to
- suffix the name of this type with `Holds` since it's a predicate primarily about `members` + `freshSelectors`
- name the conjuncts by turning this `def` into a `structure ... : Prop` definition, as in


```suggestion
/-- Rational states `members` and `freshSelectors` together satisfy the
SOS1 Big-M formulation, whose reused members are `reused` and selector bounds
are `bounds`. -/
structure PlannedSelectorFormulationHolds [Fintype ι] [DecidableEq ι]
    (reused : Finset ι) (bounds : SelectorBounds ι)
    (members freshSelectors : ι → Rat) : Prop where
  allSelectorsAreBinary :
      GenericBinaryOn Finset.univ (plannedSelector reused members freshSelectors)
  nonReusedHaveLinkedSelectors : ∀ i, i ∉ reused →
      OptionalUpperLink (bounds.upper i) (members i) (freshSelectors i) ∧
      OptionalLowerLink (bounds.lower i) (members i) (freshSelectors i)
  selectorsSumLe1 : ∑ i, plannedSelector reused members freshSelectors i ≤ 1
```

kory33-jij (commented):
I've reviewed main `theorem`s' statements and `def`s upto SOS1BigM (only `Instance/Transform/IndicatorBigM` remains to be reviewed in `OMMXProof`), and things LGTM so far.

lean/OMMXProof/Instance/Transform/IndicatorBigM/Formulation.lean
```diff
@@ -0,0 +1,110 @@
...
+-/
+
+namespace OMMXProof
+
+namespace Instance
+
+namespace IndicatorBigM
+
+/-- The upper Big-M side `f(x) + u y - u ≤ 0`, omitted when `u ≤ 0`. -/
+def UpperSide (body : State n → Rat) (trigger : Fin n) (upper : Rat)
```
kory33-jij, line 20: [imo] I am not sure if calling this an `UpperSide` is idiomatic in optimisation context. I'd suggest `UpperInequality` instead, unless you'd like rather align the terminology with the rest of ommx codebase, which appears to be using "upper/lower side".

lean/OMMXProof/Instance/Transform/IndicatorBigM/Plan.lean
```diff
@@ -0,0 +1,109 @@
...
+The selected Indicator must have a binary active-on-one trigger. A finite upper
+body bound is required for both supported body senses; equality additionally
+requires a finite lower body bound. -/
+def Valid {source : Instance n} (plan : Plan source) : Prop :=
+  source.domains plan.constraint.trigger = .binary ∧
+    plan.constraint.polarity = .activeOnOne ∧
+    plan.bodyBound.upper.IsFinite ∧
+    match plan.constraint.body.sense with
+    | .lessEqual => True
+    | .equal => plan.bodyBound.lower.IsFinite
```
kory33-jij: [imo] we should name these conditions
```suggestion
structure Valid {source : Instance n} (plan : Plan source) : Prop where
  triggerBinary : source.domains plan.constraint.trigger = .binary
  polarityActiveOnOne : plan.constraint.polarity = .activeOnOne
  upperFinite : plan.bodyBound.upper.IsFinite
  lowerFinite :
    match plan.constraint.body.sense with
    | .lessEqual => True
    | .equal => plan.bodyBound.lower.IsFinite
```

kory33-jij (approved):
I'd make two additional minor [imo] comments, but otherwise LGTM! I now have a good understanding on Indicator/SOS1 Big-M lowering operations.

lean/OMMXProof/Instance/Transform/Basic.lean
termoshtt: fixed in https://github.com/Jij-Inc/ommx/pull/1121/commits/d116717f7b7a6bb238ae0dd0574b7f00c9eef584

lean/OMMXProof/Instance/Transform/Basic.lean
termoshtt: fixed in https://github.com/Jij-Inc/ommx/pull/1121/commits/d116717f7b7a6bb238ae0dd0574b7f00c9eef584

lean/OMMXProof/Instance/Transform/SOS1BigM/Formulation.lean
termoshtt: fixed in https://github.com/Jij-Inc/ommx/pull/1121/commits/d116717f7b7a6bb238ae0dd0574b7f00c9eef584

lean/OMMXProof/Instance/Transform/IndicatorBigM/Plan.lean
termoshtt: fixed in https://github.com/Jij-Inc/ommx/pull/1121/commits/d116717f7b7a6bb238ae0dd0574b7f00c9eef584

lean/OMMXProof/Instance/Transform/IndicatorBigM/Formulation.lean
termoshtt: I keep `UpperSide` for aligning the SDK terminology.

lean/OMMXProof/Instance/Transform/SOS1BigM/Formulation.lean
termoshtt: Fixed in https://github.com/Jij-Inc/ommx/pull/1121/commits/d116717f7b7a6bb238ae0dd0574b7f00c9eef584 and https://github.com/Jij-Inc/ommx/pull/1121/commits/6b5c03c36b9554b5a09ff9fb6a977cb03f5be38d

termoshtt:
Thanks so much for the thorough review!

Issue #1111, opened 2026-07-17
RFC: Define policy-driven Adapter input preparation


termoshtt:
## Summary

Define a policy-driven OMMX workflow that automatically prepares a user-supplied
`Instance` into the concrete input accepted by an Adapter, without weakening the
direct-input contract established in #1081.

An Adapter continues to receive only an `Instance` for which
`Applicable(Adapter, input)` holds. Automatic preparation belongs to an OMMX
orchestration layer above the direct `solve()` / `sample()` calls. That layer
selects an Adapter-approved preparation path, applies it on an isolated value,
checks applicability again, invokes the Adapter, and transports or reevaluates
the output against the source when the preparation guarantee permits it.

This issue is the parent RFC and implementation workstream for that user-facing
workflow. It does not broaden any Adapter's `INPUT_CLASS`.

## Motivation

The input-class model from #1081 intentionally answers only whether an Adapter
can receive one concrete `Instance` without preparation. That boundary is
necessary, but it is not the complete user workflow.

Users commonly expect OMMX to perform suitable preparation automatically:

- lower OneHot, Indicator, or SOS1 constraints before calling a linear Adapter;
- encode bounded Integer variables before calling a Binary sampler;
- reverse an optimization sense when the target accepts only minimization;
- introduce slack variables or a penalty model when explicitly allowed.

Today these operations are fragmented. Callers prepare HiGHS inputs manually,
the OpenJij Adapter in #1087 provides an Adapter-specific `prepare()` prototype,
and `Instance.to_qubo()` is a separate mutating driver with its own fallback and
penalty defaults. There is no common selection policy, composed semantic
guarantee, result-transport contract, or durable Experiment relation between
the source and the actual Adapter input.

Restoring implicit mutation inside Adapter constructors would lose the boundary
established by #1081 and would again make it unclear which mathematical problem
an Adapter output belongs to. The automatic workflow must therefore be
convenient without becoming implicit or unauditable.

## Minimal domain vocabulary

The vocabulary from #1081 remains authoritative for Adapter input,
`InstanceClass`, membership, Adapter preconditions, and Adapter applicability.
This work adds only the following terms:

| Term | Meaning |
| --- | --- |
| source | The exact `Instance` supplied by the caller to the preparation workflow |
| Preparation | An auditable result connecting one source to the concrete Adapter input produced from it; identity preparation is allowed |
[...]

PR #1137, opened 2026-07-31
Add policy-driven SolverAdapter preparation

termoshtt:
## Summary

- add a common explicit Preparation and PreparationPolicy workflow to SolverAdapter while keeping direct solver APIs preparation-free
- migrate OpenJij existing preparation pipeline to the shared types and the Transform, policy, and decode vocabulary
- declare exact special-constraint lowering candidates for HiGHS so prepare automatically lowers Indicator, OneHot, and SOS1 inputs before final applicability validation
- document applicability, policy restriction, transformation receipts, and source-side decoding in the English and Japanese docs

## Scope

This implements the first canonical special-constraint lowering slice from #1111. It does not introduce general path search or special-constraint promotion and recovery.


termoshtt:
Closing this draft because the Preparation ownership model has been revised in #1111. The replacement will be rebuilt from origin/main around Instance.prepare(policy), with the Policy carrying the acceptable InstanceClass.

PR #1144, opened 2026-08-05
Add in-place fixed-weight penalty methods

termoshtt:
## Summary

- add `Instance::penalty_method_with_fixed_weights` for constraint-ID-keyed fixed weights
- add `Instance::uniform_penalty_method_with_fixed_weight` for one uniform fixed weight
- require keyed weights to cover exactly the active regular constraint IDs
- guarantee that `Ok(())` leaves no active regular, indicator, one-hot, or SOS1 constraints
- reject active special constraints atomically; these operations do not penalty-convert special constraints
- apply fixed weighted squares directly to the objective without creating parameter IDs or assignments
- prepare fallible objective arithmetic and regular-constraint lifecycle changes locally, then commit only after success
- leave existing `Instance` parameter assignments exactly unchanged
- record each fixed owner operation as removal provenance without transient parameter metadata
- preserve the exact `Instance` when there are no active constraints
- preserve already-removed constraints from every family
- cover the mathematical formula, ID binding, exact key domain, postcondition, identity, parameter preservation, and failure atomicity
- document the new Rust SDK operations in the 3.0 release notes

## Context

This is the next independent Rust owner-API prerequisite for the Preparation stack in #1111. Preparation can rely on successful fixed-weight penalty conversion to satisfy the no-active-constraints target membership; special constraints must be lowered before this operation. The fixed-weight APIs implement the direct numeric owner operation, while the existing consuming parameterized penalty APIs remain unchanged and separate. This PR does not introduce Preparation itself or change Python, adapters, generated artifacts, Sphinx documentation, or Lean.

Part of #1111.


termoshtt:
@codex review again

PR #1187, opened 2026-09-01
Add best-effort promotion for legacy v1 constraint hints

termoshtt:
## Summary

- Add `Instance::from_v1_bytes_with_promotion` as an explicit best-effort loader for legacy v1 OneHot and SOS1 hints, returning the decoded instance and a structured report.
- Preserve every raw hint, its original family index and wire order, and either its promoted constraint ID or original rejection error chain.
- Keep SOS1 hint-to-request conversion independently usable through `Instance::sos1_big_m_promotion_request_from_v1_hint`, returning a singleton `Sos1BigMPromotionRequest` batch map.
- Reuse the existing OneHot and SOS1 promotion APIs without changing their Plan/Apply implementations, cloning the instance, or adding rollback.

## Behavior

`Instance::from_v1_bytes` continues to ignore advisory hints and preserve the referenced regular constraints. The alternate loader parses the same base instance. Only protobuf decoding or base-instance parsing failures produce an outer `Err`; hint-specific rejections remain report data so independent valid hints can still be promoted.

OneHot hints are normalized to the existing source-ID set request. Duplicate source IDs share one promotion result, and the legacy member list remains advisory: membership comes from the regular row. The existing `From<&v1::OneHot>` conversion remains independently usable.

SOS1 conversion reconstructs member, selector, and link roles from the original regular rows before any promotion:

- Identical normalized claims for the same cardinality ID are merged into one request, regardless of member/link wire order.
- Different converted claims for that ID reject every occurrence in the group; no claim wins by input order.
- A conversion failure rejects only that hint and does not block other hints.

Each raw occurrence still receives its own report entry. Duplicate normalized requests share the allocated target ID or an `Arc<ommx::Error>`, preserving the original diagnostic chain without reconstructing an error from its message.

OneHot promotion completes first. The SOS1 batch then validates against the resulting instance, including fresh-selector interference with promoted OneHot constraints. Each family owns its borrowed plan and infallible Apply; the loader introduces no second plan and reports both successful and rejected owner results. Remaining `expect` calls assert only singleton-conversion and exact result-key contracts, not successful promotion.

The supplied `ATol` is used for SOS1 Big-M verification. OneHot retains its existing exact-binary feasible-set guarantee without promising identical finite-tolerance feasibility classification.

## Scope

This is a Rust SDK addition, including public API documentation and Rust release notes. It does not change the protobuf schema, Python SDK, Lean formalization, OneHot promotion implementation, or SOS1 batch planning/application.


termoshtt:
@codex review again

